PR #1, opened 2020-08-13
Initial Commit

ZER0:


.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+os:
+  - linux
+
+matrix:
+  include:
+    - rust: nightly
+      after_script:
+          # Send a notification to the Dusk build Status Telegram channel once the CI build completes
+        - bash <(curl -s https://raw.githubusercontent.com/dusk-network/tools/master/bash/telegram_ci_notifications.sh)
+
```
CPerezz, line 16: We should also publish the docs. But it's fine atm

Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "phoenix-core"
+version = "0.1.0"
+authors = ["zer0 <matteo@dusk.network>", "Victor Lopez <victor@dusk.network"]
+edition = "2018"
+
+[dependencies]
+rand = "^0.7"
+kelvin = "0.18"
+poseidon252 = { git = "https://github.com/dusk-network/Poseidon252", tag = "v0.6.0" }
```
CPerezz, line 10: Should be `tag = v0.6.1`

src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+pub mod note;
+
+pub use error::Error;
+pub use note::{Note, NoteType};
+
+use dusk_plonk::bls12_381::Scalar as BlsScalar;
+use dusk_plonk::jubjub::{
+    AffinePoint as JubJubAffine, ExtendedPoint as JubJubExtended,
+    Fr as JubJubScalar,
+};
```
CPerezz, line 18: Too verbose.

Use:
```rust
use dusk_plonk::prelude::*;
```

src/macros.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_export]
```
CPerezz, line 1: A small description comment would be nice.

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+use dusk_plonk::jubjub::{dhke, GENERATOR_EXTENDED, GENERATOR_NUMS_EXTENDED};
+use poseidon252::cipher::PoseidonCipher;
+use poseidon252::sponge::sponge::sponge_hash;
+use rand::Rng;
+
+use crate::{
+    chunk_of, BlsScalar, Error, JubJubAffine, JubJubExtended, JubJubScalar,
+};
+
+const ENCRYPTED_DATA_SIZE: usize = 96;
```
CPerezz, line 19: If that's refering to the `PoseidonCipher` I think the constant should be defined in `Poseidon252` repo. Not here.

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+
+/// A note that does not encrypt its value
+#[derive(Clone, Copy)]
+pub struct Note {
+    note_type: NoteType,
+    value_commitment: JubJubExtended,
+    nonce: JubJubScalar,
+    stealth_address: StealthAddress,
+    pos: u64,
+    encrypted_data: [u8; ENCRYPTED_DATA_SIZE],
```
CPerezz, line 55: Why not use `PoseidonCipher` here? See: https://github.com/dusk-network/Poseidon252/blob/17329418740028b571752e936b190a1f15f94c5a/src/cipher/cipher.rs#L84

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+            .field("nonce", &self.nonce)
+            .field("stealth_address", &self.stealth_address)
+            .field("pos", &self.pos)
+            .field(
+                "encrypted_data",
+                &format_args!("{:?}", &self.encrypted_data),
+            )
+            .finish()
+    }
+}
```
CPerezz, line 87: We have it. So this should be removed or the comment should not be there.

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Creates a new obfuscated note
+    pub fn obfuscated(psk: &PublicSpendKey, value: u64) -> Self {
+        Self::new(NoteType::Obfuscated, psk, value)
+    }
+
+    /// Create a new phoenix output note without inner randomness
+    pub fn deterministic(
+        note_type: NoteType,
+        r: &JubJubScalar,
+        nonce: JubJubScalar,
```
CPerezz, line 176: If that's the `PoseidonCipher` nonce. The type is not correct. It should be `BlsScalar` as per https://github.com/dusk-network/Poseidon252/blob/17329418740028b571752e936b190a1f15f94c5a/src/cipher/cipher.rs#L110

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+        value: u64,
+        blinding_factor: JubJubScalar,
+    ) -> Self {
+        let stealth_address = psk.gen_stealth_address(r);
+
+        let value_commitment = JubJubScalar::from(value);
+        let value_commitment = JubJubExtended::from(JubJubAffine::from(
+            &(GENERATOR_EXTENDED * value_commitment)
+                + &(GENERATOR_NUMS_EXTENDED * blinding_factor),
+        ));
```
CPerezz, line 187: No need to do this conversions. 
```rust
let value_commitment = (GENERATOR_EXTENDED * value_commitment)
                + (GENERATOR_NUMS_EXTENDED * blinding_factor);
```

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+        cipher
+            .write(&self.encrypted_data[..])
+            .expect("Cannot write the PoseidonCipher");
+
+        let data = cipher.decrypt(&shared_secret, &nonce);
+
+        // TODO: need to figure out what we want to do in case the decryption
+        // fails
+        return if data.is_err() {
+            (0u64, JubJubScalar::zero())
```
CPerezz, line 248: We should simply return the error IMO. If the decryption fails you don't have anything to do with the Note.

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Return the position of the note on the tree.
+    pub fn pos(&self) -> u64 {
+        self.pos
+    }
+
+    /// Set the position of the note on the tree.
+    /// This, naturally, won't reflect immediatelly on the data storage
+    pub fn set_pos(&mut self, pos: u64) {
+        self.pos = pos;
+    }
```
CPerezz, line 296: Will this get the Index returned by the storage call from kelvin and store it?
If that's the case, this looks correct to me. Otherways doesn't since you can't decide where kelvin stores.

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+            }
+            NoteType::Obfuscated if vk.is_some() => {
+                let (value, _) = self.decrypt_data(vk.unwrap());
+                value
+            }
+            _ => {
+                let mut rng = rand::thread_rng();
+                let value: u64 = rng.gen();
+                value
+            }
```
CPerezz, line 325: Shouldn't we fail here?

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        Ok(Note {
+            note_type,
+            value_commitment,
+            nonce,
+            stealth_address,
+            pos,
+            encrypted_data,
+        })
+    }
```
CPerezz, line 393: Shouldn't we add also the `Into<StorageScalar>` trait impl here?

.travis.yml
ZER0: I copied one of the `.travis.yml` we have in our repo, what would a more recent one?
Or, should we already move out from `travis` also here and move to github actions? (In that case, I will file a follow up issue)

src/note.rs
ZER0: First of all, because when this was implemented the first time we didn't have `PoseidonChiper` yet – that also why the `ECRYPTED_DATA_SIZE` is defined here. But once the `PoseidonChiper` was implemented, it was because the internals. It would be better if `PoseidonCipher` would use as internals bytes instead of `BlsScalar` – if we want to have this representation on contracts as well – but before changing either poseidon or phoenix-core it's better to cross that bridge and see what's the best approach.

src/note.rs
ZER0: No, we don't.
Basically we need to remove that once we have the "proper" encrypted data type (see comment above) and we don't have it yet – since it would be better waiting, otherwise we've 50% chance to re-write the wrong part of the code.

src/note.rs
ZER0: `phoenix-core` doesn't deal with `BlsScalar` directly (in the type structures I mean) but with JubJub's instead. The fact that Poseidon only deals with `BlsScalar` is an implementation details – meaning: from the point of the phoenix-core's consumer, Poseidon is a black box.

Having Nonce as BlsScalar in a public struct is an anomaly in the design, and leaking an internal type.

I talked with Togh and he doesn't care if it's JubJub or BlsScalar, and he's going to update the docs.

It's like the `value`, basically, that is passed and stored as u64 but then is converted internally as `BlsScalar: the consumers doesn't need to know that, they shouldn't care about that.

The alternative would be have the Nonce as simple sequence of 32 bytes (that "represent" a BlsScalar), from a design point of view is also valid.

src/note.rs
CPerezz: @khovratovich @vlopes11 any suggestions?

src/note.rs
CPerezz: kk. Fair enough

src/note.rs
CPerezz: No, I'm happy then to have it as a jubjub or bls Scalar.

.travis.yml
CPerezz: I don't have any problem staying with travis. Specially because the builds & tests here should not take too long since there's not any ZK stuff going on.

The block you're missing for the docs upload is the following:
```yaml
# Upload docs
  - |
      if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_RUST_VERSION" == "nightly" && "$TRAVIS_PULL_REQUEST" = "false" && "$TRAVIS_BRANCH" == "master" ]]; then
        cargo rustdoc --features nightly -- --html-in-header katex-header.html &&
        echo "<meta http-equiv=refresh content=0;url=dusk_plonk/index.html>" > target/doc/index.html &&
        git clone https://github.com/davisp/ghp-import.git &&
        ./ghp-import/ghp_import.py -n -p -f -m "Documentation upload" -r https://"$GH_TOKEN"@github.com/"$TRAVIS_REPO_SLUG.git" target/doc &&
        echo "Uploaded documentation"
      fi
```

i usually have it inside an `after_success:` block

CPerezz: Replace `dusk_plonk` by the name of the crate and it should work without any problems.

Cargo.toml
vlopes11: Was it updated?

src/note.rs
vlopes11: We have on the same structure a `JubJubScalar`, so in this case we do not have only bytes. Why does this apply?

src/note.rs
```diff
@@ -0,0 +1,394 @@
...
+                    &nonce,
+                );
+
+                // TODO: replace the internals of PoseidonCipher with a plain
+                // array.
+                // Adding an infallible `to_bytes` method too.
+                let mut encrypted_data = [0u8; 96];
+                cipher
+                    .read_exact(&mut encrypted_data)
+                    .expect("Cannot encode PoseidonCipher");
```
vlopes11, line 214: If the operation is fallible then we should work with `Result`. Since the `Note` contains other scalars, then we have no strict requirement to use only bytes.

.travis.yml
ZER0: I might just use the Makefile for that, in this case.

src/lib.rs
ZER0: Solving this because we discussed about when to use `prelude` and when not (basically for `dusk_plonk` prelude is necessary only when we're going to use the composer / plonk, not when we want to access only to the re-exported crates).

For future reference, we might change this approach in the future (but is not something we're going to do atm).

src/macros.rs
ZER0: I will add it, I'm also not sure if I want to keep it actually, since now (after the refactoring) is used way less.

src/note.rs
ZER0: As discussed, it was needed because the hashing. Now we're using `to_hash_inputs()` and there is no need to such conversion here.

src/note.rs
ZER0: Basically yes. This was used by the function `store_unspent_note` in the original code, and was set by the DB (Kelvin).

It's not used yet because that part is still missing (either `Rusk` that will take care of it or `phoenix-core` once is clear how the PoseidonTree are working from `Rusk-VM`).

src/note.rs
ZER0: I added in the error refactoring mentioned previously also a failing here, since now we're returning a `Result`.

src/note.rs
ZER0: I was thinking to add in a follow up PR, in order to land this – since the `StorageScalar` is also part of the `PoseidonTree` functionality that is still missing – but wen can also add the implementation directly in the Initial Commit.
You're the reviewer, let me know what's your preference.

Personally I'm more about progressive enhancement unless is a key function that prevent everyone to contribute to the repo.

src/note.rs
ZER0: In order to use directly the `PoseidonCipher` as `encrypted_data`, that was my initial idea, we need few small changes in the `poseidon` repo, otherwise we might complicate a bit the code here.

Since `encrypted_data` is encapsulated and not exposed, I don't think it's a blocker for the landing. I will therefore proceed to merge this PR, open a PR for the `poseidon252` repo, and once the changes on the `poseidon252` are landed proceed to change the internals in the `Note`.

src/note.rs
CPerezz: If you want to do it on a follw-ip PR, feel free to add the issue now.

CPerezz (approved):
LGTM_ACK

ZER0:
As discussed I filed issue #8 (that is a blocker for issue #5) so we can merge the `phoenix-core`'s initial commit and unblock the works on circuits.

PR #13, opened 2020-09-28
Add a method to fully reconstruct a Note

### LINKED ISSUE #10: Allow a consumer of the library to specify all fields of a `Note` directly
This functionality is needed in order to successfully decode a `Note` from a GRPC connection. Currently, the library only offers the option to supply some "pre-image" variables, which are then turned into the proper fields by using some function.

The stealth address is a good example of this: in `Note::deterministic`, you can only supply the public spend key and the random scalar that make up this stealth address - variables that you will not have if you are just decoding a Note from GRPC.

jules:
This method is specifically needed in order
to be able to decode Notes from protocol
buffers.

Fixes #10

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -228,7 +228,7 @@ impl Note {
     }
 
     /// Reconstruct a Note from its individual fields.
-    pub fn reconstruct(
+    pub fn from_fields(
         note_type: NoteType,
         value_commitment: JubJubExtended,
         nonce: JubJubScalar,


src/note.rs
```diff
@@ -226,6 +227,25 @@ impl Note {
         }
     }
 
+    /// Reconstruct a Note from its individual fields.
+    pub fn reconstruct(
```
CPerezz, line 231: IDK if the name is really telling what we're doing here. I think methods like `new` or `from_fields` would fit better.

julesdesmit: Fair enough. `new` is already taken so I was unsure of what name to pick - maybe `from_fields` is better.

CPerezz (approved):
LGTM.

@ZER0 ?

src/note.rs
```diff
@@ -226,6 +227,25 @@ impl Note {
...
+    ) -> Self {
+        Note {
+            note_type,
+            value_commitment,
+            nonce,
+            stealth_address,
+            pos,
+            encrypted_data,
+        }
+    }
```
ZER0, line 247: I would close this PR, a method such this one makes no sense. If you need to create a struct passing all the fields means you basically creates the struct directly, so:
```rust
let my_note = Note {
            note_type,
            value_commitment,
            nonce,
            stealth_address,
            pos,
            encrypted_data,
        };
```

ZER0 (changes requested):

src/note.rs
ZER0: To be clear, this won't be the approach I would take, since it would means everything would be in write / read mode for the consumer. In term of design API we need to understand why we need specifically this, and figure out an alternative.

julesdesmit: This approach also would not work, since none of the fields are public. Hence why I have the method.

I think the issue/commit explains why we need it (decoding from protocol buffers), without this I'd need to write a lot more code in order to actually decode a note.

julesdesmit: After a brief discussion on call, the conclusion was to not expose any of these fields directly and instead using the `Read/Write` traits to write all the data onto a newly decoded `Note`.

PR #12, opened 2020-09-28
Add a function to return encrypted data of a Note

jules:
This allows consumers of the library to
read all of the fields contained within
the Note struct, which is necessary in order
to properly encode Notes to protocol buffers.

Fixes #11

PR #14, opened 2020-09-29
Revert "Add a function to return encrypted data of a Note"

julesdesmit:
After a brief discussion, we decided instead not to change the API of phoenix-core, but instead using the `Read/Write` traits in order to read what is not accessible directly.

Reverts dusk-network/phoenix-core#12

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -1,6 +1,5 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
-// Licensed under the MPL 2.0 license. See LICENSE file in the project root for
-// details.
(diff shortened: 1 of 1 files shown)


ZER0 (approved):
To give a bit of context, the API shouldn't change because the wrongly way we're using gRPC – or at least, it should change in a specific way for the precise way we're using gRPC, since the approach we use gRPC is wrong and might change in the future.
Since `Note` already provides `Read` and `Write` method to serialize / deserialize a Note without accessing directly to the internals, we should use that.

Also because we might decide to simplify in the close future the sending and receiving of datas via gRPC.

Issue #8, opened 2020-08-26
Use `PoseidonCipher` as part of the `Note` structure


ZER0:
Currently `PoseidonCipher` is used only as meaning to decrypt / encrypt the `encrypted_data`, that is an array of 96 bytes.

We should probably use it as type instead of the `encrypted_data` itself.

That might require a bit of changes on dusk-network/poseidon252 repo as well (e.g. for consistency in the code we might need a `from_bytes` and `to_bytes` method, to simplify the encoding / decoding of `Note` type)

PR #17, opened 2020-09-30
Use `PoseidonCipher` as type for `encrypted_data`

ZER0:
Previously `PoseidonCipher` was used only to decrypt / encrypt, instead
of be used directly as type in the structures.

Resolves: #8


CPerezz (approved):
LGTM

PR #21, opened 2020-10-03
Use as default Note's position `u64::MAX` instead of `0`

ZER0:
Since `0` is a valid position inside a tree, a `Note` shouldn't have it
as default value. We used `u64::MAX` instead since this value is not
valid as tree's index.

- Fix few nits (comments, unnecessary unwraps)


julesdesmit (approved):
LGTM

PR #28, opened 2020-10-19
Amend preimage 

### LINKED ISSUE #27: Amend hash method
The existing hashing method, `pub fn hash`,  for the note doesn't hash enough field inputs. As a result, ownership of the note cannot be proven by anybody except the one spending the note. 

The current method hashes the note position `pos`, the public spend key `pk_r`, and the value commitment.

The additional fields to be hashed will be the randomness `R`, the note type, the nonce and the scalars within the encrypted data.

LukePearson1:
Closes #27

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -217,7 +217,7 @@ impl Note {
         let R = self.stealth_address().R().to_hash_inputs();
         let cipher = self.encrypted_data.cipher();
 
-        // We assume cipher contains three scalars, 
+        // We assume cipher contains three scalars,
         // this could change in the future.
         sponge_hash(&[
             BlsScalar::from(self.note_type as u64),


src/note.rs
```diff
@@ -209,18 +209,26 @@ impl Note {
...
             value_commitment[0],
             value_commitment[1],
-            BlsScalar::from(self.pos()),
+            BlsScalar::from(self.nonce),
             pk_r[0],
             pk_r[1],
+            R[0],
+            R[1],
+            BlsScalar::from(self.pos()),
+            self.encrypted_data.cipher()[0],
```
ZER0, line 229: Add the cipher as separate variable, something like:
```rust
         let value_commitment = self.value_commitment().to_hash_inputs();
         let pk_r = self.stealth_address().pk_r().to_hash_inputs();
         let R = self.stealth_address().R().to_hash_inputs();
         let cipher = self.encrypted_data.cipher();
```

and then:

            R[1],
            BlsScalar::from(self.pos()),
            cipher[0],
            cipher[1],
            cipher[2]
```

Also, add a comment where here we're assuming `cipher` has length 3. We could add an assert but it would be an additional check currently we don't need, and I've in mind an alternative solution.

**Review by ZER0 — CHANGES_REQUESTED** [review]

File: src/note.rs [hunk 2]
```diff
@@ -209,18 +209,31 @@ impl Note {
...
+    /// Return a hash represented by `H(note_type, value_commitment,
+    /// H(StealthAddress), pos, encrypted_data)
     pub fn hash(&self) -> BlsScalar {
         let value_commitment = self.value_commitment().to_hash_inputs();
         let pk_r = self.stealth_address().pk_r().to_hash_inputs();
+        let R = self.stealth_address().R().to_hash_inputs();
+        // The cipher is made being used differently to the
+        // other variables as we here enforce the PoseidonCipher
+        // wihtin each note to contain three scalars. This is a
+        // subsitution for an assert! method.
```
ZER0, line 221: I would remove the comment from here. Just add before the `sponge_hash` something like "// we assume cipher contains three scalars, this could change" or similar.

ZER0 (approved):
Approved with the comment slightly changed.

PR #30, opened 2020-11-02
Return cipher as array

### LINKED ISSUE #29: Create a method for retrieving the encrypted data from the note
Currently only within the phoenix core library can the ciphers be called for the hashing method. 

For the method to be used elsewhere, we need to be able to call these fields when the library is exported.

LukePearson1:
Here we implement a function for returning the encrypted data as an array of BlsScalars

This closes #29

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -261,10 +261,9 @@ impl Note {
         &self.value_commitment
     }
 
-    /// Returns the encryped data from the note as an array of
-    /// BlsScalar's
-    pub fn cipher(&self) -> [BlsScalar; CIPHER_SIZE] {
-        *self.encrypted_data.cipher()
+    /// Returns the cipher of the encrypted data
+    pub fn cipher(&self) -> &[BlsScalar; CIPHER_SIZE] {
+        self.encrypted_data.cipher()
     }
 
     /// Attempt to decrypt the note value provided a [`ViewKey`]. Always


CPerezz (changes requested):
The code addition is fine, but you'll need to bump the patch version of the crate otherways you'll not be able to use it.

src/note.rs
```diff
@@ -261,6 +261,12 @@ impl Note {
         &self.value_commitment
     }
 
+    /// Returns the encryped data from the note as an array of
+    /// BlsScalar's
```
ZER0, line 265: This should be more like `Returns the encrypted data's cipher` or something like that.

src/note.rs
```diff
@@ -261,6 +261,12 @@ impl Note {
...
+    pub fn cipher(&self) -> [BlsScalar; CIPHER_SIZE] {
+        *self.encrypted_data.cipher()
```
ZER0, line 267: This should match the signature of the actual `cipher` method that we're proxying, so it should be: 
```rust
pub fn cipher(&self) -> &[BlsScalar; CIPHER_SIZE] {
    self.encrypted_data.cipher()
}
```

ZER0 (changes requested):
Small nits, mostly about the function's signature

CPerezz (approved):
LGTM. Remember to release a new version in GH.

PR #32, opened 2020-11-02
Update CI to be Canon-compatible

### LINKED ISSUE #31: Adapt CI to be Canon compatible
Since we have `Canon` in the deps of the crate's deps, we need to adapt the CI to be compatible with it.

`Canon` usage requires `min_const_generics` feature enabled, and therefore, nightly toolchain. So we need to remove `stable` toolchain tests from there.

This will unblock #29

CPerezz:
We need a CI that makes sense with the latest
additions of `Canon` trait that we added to our
stack as stated in #31.

So this PR modifies the CI to:
- `stable` toolchain is no longer tested/supported.
- `cargo check` is done in nightly toolchain also.
- `rustfmt` checks are done with a fixed nightly
version due to the chances of a new nightly toolchain
appearing without rustfmt compiling are quite high.

Closes #31 and unblocks #29

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -29,26 +29,11 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
       - uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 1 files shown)


vlopes11 (approved):
LGTM!

PR #37, opened 2020-11-20
Canon & no-std phoenix core

vlopes11:


Cargo.toml
```diff
@@ -5,16 +5,25 @@ authors = ["zer0 <matteo@dusk.network>", "Victor Lopez <victor@dusk.network"]
...
-kelvin = "0.19"
-poseidon252 = { git = "https://github.com/dusk-network/Poseidon252", tag = "v0.10.0" }
-dusk-pki = { git = "https://github.com/dusk-network/dusk-pki", tag = "v0.3.0" }
-thiserror = "1.0"
-dusk-plonk = { version = "0.3.3", features = ["trace-print"] }
+rand_core = "0.5.1"
+dusk-bls12_381 = {version = "0.3", default-features = false}
+dusk-jubjub = {version = "0.5", default-features = false}
+poseidon252 = {git = "https://github.com/dusk-network/Poseidon252", tag = "v0.14.1", default-features = false}
+dusk-pki = {git = "https://github.com/dusk-network/dusk-pki", tag = "v0.4.0", default-features = false}
```
CPerezz, line 12: This is not using the latest Poseidon252 version. We should bump PKI first as I suggested so that we don't have any kind of duplicated deps.

CPerezz (changes requested):
Dependency questions.

Cargo.toml
vlopes11: Fixed

CPerezz (approved):
LGTM

PR #39, opened 2020-12-04
Add to/from bytes impls for Fee & Crossover

### LINKED ISSUE #38: Bring back read/write substitutive methods for Crossover & Fee.
In the latest release, the IO impls of read/write were removed but no substitute was added.

So we need a to/from bytes impl for these structures.

CPerezz:
Closes #38

### DIFF
--- a/src/crossover.rs
+++ b/src/crossover.rs
@@ -6,13 +6,15 @@
 
 //! Fee module contains the logic related to `Crossover` structure
 
-use crate::{BlsScalar, JubJubExtended, JubJubScalar};
+use crate::{BlsScalar, Error, JubJubExtended, JubJubScalar};
 
 #[cfg(feature = "canon")]
 use canonical::Canon;
 #[cfg(feature = "canon")]
 use canonical_derive::Canon;
 
+use dusk_jubjub::JubJubAffine;
+use dusk_pki::jubjub_decode;
 use poseidon252::cipher::PoseidonCipher;
 use poseidon252::sponge::hash;
 
@@ -34,6 +36,11 @@ impl PartialEq for Crossover {
 impl Eq for Crossover {}
 
 impl Crossover {
+    /// Returns the serialized size of the Crossover
+    pub const fn serialized_size() -> usize {
+        32 * 2 + PoseidonCipher::cipher_size_bytes()
+    }
+
     /// Returns a hash represented by `H(value_commitment)`
     pub fn hash(&self) -> BlsScalar {
         let value_commitment = self.value_commitment().to_hash_inputs();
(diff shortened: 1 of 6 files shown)


src/crossover.rs
```diff
@@ -34,6 +36,11 @@ impl PartialEq for Crossover {
 impl Eq for Crossover {}
 
 impl Crossover {
+    /// Returns the serialized size of the Crossover
+    pub const fn serialized_size() -> usize {
+        32 * 2 + PoseidonCipher::cipher_size_bytes()
```
ZER0, line 41: We should rename it `serialized_size` to be consistent, since it's the same method used also for `from_bytes` and `to_bytes` for `PoseidonCipher`.

I open another issue for that: https://github.com/dusk-network/Poseidon252/issues/102

src/crossover.rs
```diff
@@ -55,4 +62,60 @@ impl Crossover {
...
+        assert_eq!(n, Crossover::serialized_size());
+
+        buf
+    }
+
+    /// Attempts to convert a byte representation of a note into a `Note`,
+    /// failing if the input is invalid
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        if bytes.len() < Crossover::serialized_size() {
+            return Err(Error::InvalidCrossoverConversion);
```
ZER0, line 92: It's okay because you've been consistent within the repos, but I think we start to have more clear errors for such scenarios: "InvalidXXXConversion" doesn't tell why. Here the error is that we don't have enough bytes into the buffer for actually even trying a conversion.

You can search for similar code in our repos, and get an inspiration for what I mean, we might want to refactor this also in the future.

src/crossover.rs
```diff
@@ -55,4 +62,60 @@ impl Crossover {
...
+        n += 32;
+
+        buf[n..n + 32].copy_from_slice(&self.nonce.to_bytes()[..]);
+        n += 32;
+
+        buf[n..n + PoseidonCipher::cipher_size_bytes()]
+            .copy_from_slice(&self.encrypted_data.to_bytes()[..]);
+        n += PoseidonCipher::cipher_size_bytes();
+
+        assert_eq!(n, Crossover::serialized_size());
```
vlopes11, line 83: We can't have assertions in the middle of the code.

There is a better option `debug_assert` that will run these assertions in debug mode(non-optimized compilation)

src/fee.rs
```diff
@@ -97,6 +102,58 @@ impl Fee {
...
+        buf[n..n + 8].copy_from_slice(&self.gas_limit.to_le_bytes()[..]);
+        n += 8;
+
+        buf[n..n + 8].copy_from_slice(&self.gas_price.to_le_bytes()[..]);
+        n += 8;
+
+        buf[n..n + 64].copy_from_slice(&self.stealth_address.to_bytes()[..]);
+        n += 64;
+
+        assert_eq!(n, Fee::serialized_size());
```
vlopes11, line 120: Same as above

vlopes11 (changes requested):
Small nit, the remainder is LGTM!

ZER0 (approved):
LGTM, but check out the comments anyway.

src/crossover.rs
ZER0: It depends, sometimes is actually helps the compiler to optimize code; but if it's used at the beginning, not in this case of course. Agreed it should be done a `debug_assert` or skipped at all.

ZER0 (approved):

src/crossover.rs
CPerezz: I thought the same while coding it. But the assertions were already in other fn's. So I basically went on the same direction. Will use `debug_assert!` as I did with zerocaf back then.

src/crossover.rs
CPerezz: We can file an issue for it. It would be nice to be consistent although I don't think we should give prio to this atm.

src/fee.rs
```diff
@@ -97,6 +102,58 @@ impl Fee {
...
+
+        one_u64.copy_from_slice(&bytes[n..n + 8]);
+        let gas_price = u64::from_le_bytes(one_u64);
+        n += 8;
+
+        one_stealth_addr.copy_from_slice(&bytes[n..n + 64]);
+        let stealth_address = StealthAddress::from_bytes(&one_stealth_addr)?;
+        n += 64;
+
+        assert_eq!(n, Fee::serialized_size());
```
vlopes11, line 149: Check `assert_eq`

vlopes11 (changes requested):
Small nit

src/fee.rs
CPerezz: MB sorry.

vlopes11 (approved):
LGTM!

PR #42, opened 2021-01-06
Fixes #41- Wrong value commitment for transp notes

### LINKED ISSUE #41: [Bug] Value Commitment wrong for transparent notes
## Problem

If the note is transparent, then the blinding factor is always returned as `JubJubScalar::zero()`:
https://github.com/dusk-network/phoenix-core/blob/c34580bf6e24df4601d5313b8b8c418a06d9d7ad/src/note.rs#L271

However, the value commitment is calculated with a random blinding factor.

1) Fetch a random number generator
https://github.com/dusk-network/phoenix-core/blob/c34580bf6e24df4601d5313b8b8c418a06d9d7ad/src/note.rs#L92
2) Use it to produce the blinding factor
https://github.com/dusk-network/phoenix-core/blob/c34580bf6e24df4601d5313b8b8c418a06d9d7ad/src/note.rs#L86
3) Calculate the value commitment with this random blinding factor
https://github.com/dusk-network/phoenix-core/blob/c34580bf6e24df4601d5313b8b8c418a06d9d7ad/src/note.rs#L122

Later on, the blinding factor is hardcoded as zero:
https://github.com/dusk-network/phoenix-core/blob/c34580bf6e24df4601d5313b8b8c418a06d9d7ad/src/note.rs#L130

Hence, it is impossible to reconstruct the value commitment inside of a circuit because the blinding factor is lost. Some workaround solutions are possible, but they will increase the number of gates of the zk circuits, and this is undesirable.

## Proposed solution

The value commitment is not required for transparent notes. To save storage, hardcode the blinding factor of transparent notes to zero before the commitment value is calculated.

This way the value commitment can be reconstructed in any circuit and also there is no need to store neither the blinding factor nor the value commitment in the blockchain.

vlopes11:
Fixes #41

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -83,7 +83,15 @@ impl Note {
     ) -> Self {
         let r = JubJubScalar::random(rng);
         let nonce = JubJubScalar::random(rng);
-        let blinding_factor = JubJubScalar::random(rng);
+
+        // Blinding factor and value commitment are open for transparent note
+        // In order to save storage, these may not be stored and should be
+        // hardcoded for an eventual proof of knowledge of the
+        // commitment
+        let blinding_factor = match note_type {
+            NoteType::Transparent => JubJubScalar::zero(),
+            NoteType::Obfuscated => JubJubScalar::random(rng),
+        };
 
         Self::deterministic(note_type, &r, nonce, psk, value, blinding_factor)
     }
@@ -113,13 +121,9 @@ impl Note {
         nonce: JubJubScalar,
         psk: &PublicSpendKey,
         value: u64,
-        mut blinding_factor: JubJubScalar,
+        blinding_factor: JubJubScalar,
     ) -> Self {
         let stealth_address = psk.gen_stealth_address(r);
-
-        if let NoteType::Transparent = note_type {
-            blinding_factor = JubJubScalar::zero();
-        }
 
         let value_commitment = JubJubScalar::from(value);
         let value_commitment = (GENERATOR_EXTENDED * value_commitment)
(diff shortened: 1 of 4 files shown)


src/note.rs
```diff
@@ -113,10 +113,14 @@ impl Note {
         nonce: JubJubScalar,
         psk: &PublicSpendKey,
         value: u64,
-        blinding_factor: JubJubScalar,
+        mut blinding_factor: JubJubScalar,
```
ZER0, line 116: This shouldn't be mutable.
The `deterministic` function should accept the argument as is. We might have a `debug_assert!` to ensure the `blinding_factor` is actually zero, but that's the only thing we want to perform inside the `deterministic` method.

src/note.rs
```diff
@@ -113,10 +113,14 @@ impl Note {
...
     ) -> Self {
         let stealth_address = psk.gen_stealth_address(r);
 
+        if let NoteType::Transparent = note_type {
+            blinding_factor = JubJubScalar::zero();
+        }
```
ZER0, line 122: This should be therefore removed (as mentioned, a debug assertion might be useful instead).

The part where you want to put this code is instead in `new`:
```rust
        rng: &mut R,
        note_type: NoteType,
        psk: &PublicSpendKey,
        value: u64,
    ) -> Self {
        let r = JubJubScalar::random(rng);
        let nonce = JubJubScalar::random(rng);
        let blinding_factor = if let NoteType::Transparent = note_type {
            JubJubScalar::zero()
        } else {
            JubJubScalar::random(rng)
        };

        Self::deterministic(note_type, &r, nonce, psk, value, blinding_factor)
    }
```

src/note.rs
vlopes11: Good point, fixed

src/note.rs
vlopes11: Fixed

ZER0 (approved):
LGTM

PR #45, opened 2021-01-07
Crossover/Fee blinding factor

vlopes11:
### Blinding factor returned when create transp note

The blinding factor is a required return of the obfuscated note creation.

To maintain API consistency, the creation of the transparent note should return the blinding factor as well.

It happens to be constant zero. But, to keep this logic decoupled from the consumers of this library, and if ever this constant changes, its logic is now granted to be defined only inside of phoenix-core

###  Blinder should be returned when create obfuscated

Note::new is not part of the API because it is expected the note is created either via `Note::transparent` or `Note::obfuscated`.

The blinding factor must be returned when the note is created as obfuscated because the creator of an eventual output will not be able to decrypt it according to the properties of asymmetric encryption


vlopes11:
Closed in favor of #46

PR #46, opened 2021-01-07
Blinding factor as parameter of Note::obfuscated

vlopes11:
The blinding factor is required to prove the knowledge of the value commitment of the note.

If the blinding factor is not provided by the note creator, the latter will not be able to fetch this information, and consequentially create a zk proof of knowledge of the value commitment, without the view key of the owner of the note.

This is not desirable since a note should be created only with the public key of the receiver.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.6.0] - 07-01-21
 ### Changed
 - Blinding factor provided to create obfuscated notes
-- Transparent note with constant blinding factor
 
 ## [0.5.1] - 06-01-21
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


CPerezz (approved):
LGTM. But we should wait for releasing in crates.io or GH until jubjub and others are bumped! :)

CHANGELOG.md
```diff
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.6.0] - 07-01-21
+### Changed
+- Blinding factor provided to create obfuscated notes
+- Transparent note with constant blinding factor
```
ZER0, line 13: This is not necessary to add in the changelog, since as discussed we don't expose it in any way, and does not affect the API or the internal behavior, so it shouldn't mentioned in the changelog. It's just an implementation detail.

ZER0 (changes requested):
LGTM, since we already discussed the changes over chat and video call I have nothing to add.

Just a small nits to remove the line about the implementation detail from the changelog since the constant is never exposed, it's just used internally.

CHANGELOG.md
vlopes11: Fixed

ZER0 (approved):
LGTM

PR #66, opened 2021-02-19
Closes #65 - Fee - Out of gas exceptions

### LINKED ISSUE #65: Fee - out of gas exception
The generation of a remainder should fail if it is out of gas, and should return `None` if the gas limit = gas consumed.

Currently, there is an encapsulated workaround to avoid overflows with the assumption that all required checks are executed outside the API.

This assumption is not based on concrete restrictions and could case misuse of the API.

#### Proposed solution

Change the signature of `Fee::gen_remainder(&self, gas_consumed: u64) -> Remainder` to `Fee::gen_remainder(&self, gas_consumed: u64) -> Result<Option<Remainder>, _>`.

Throw `Err(OutOfGas)` if the gas limit is not enough.
Return `None` if `gas limit == gas consumed`

vlopes11:
Closes #65 

Prior to this commit, no gas limit or gas consumed validations were
performed. This can lead to several bugs in phoenix implementations, and
there is no reason to not perform the checks in this library since it
contains all the required data for every validation.

Also, a consumed fee note will need to be created from a given stealth
address to a block generator. Prior to this commit, the API will not
allow note creation with a specific stealth address without the
knowledge of `r`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.10.0] - 2021-02-19
+
+### Changed
+
+- Fee - Out of gas exceptions [#65](https://github.com/dusk-network/phoenix-core/issues/65)
+- Remainder - Generate note from consumed gas
+- Note - Create transparent note from stealth address
+
 ## [0.9.1] - 2021-02-11
 
 ### Changed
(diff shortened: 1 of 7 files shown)


CPerezz (approved):
LGTM. But @ZER0 is the one who has worked with you the most on this lib. So although I'm ok with the changes, I would like to see if he has anything in respect of the new logic. 

In case he is ok just with my review, feel free to merge!

PR #70, opened 2021-04-14
crossover: use all attributes as hash inputs

### LINKED ISSUE #69: crossover: use all attributes as hash inputs
Currently, the hash function is using only the value commitment as input. This will effectively be a unique representation of the crossover considering the value commitment is a composed by the value and the blinder; and so is the encrypted data.

But,the hash representation should reflect all attributes - even the redundant ones - to avoid any malleability of the components and increase the overall security of the system.

vlopes11:
The hash representation of the crossover should reflect all the
attributes.

Even if the crossover is composed only by `value` and `blinder`, and
both are present in the value commitment, its hash representation should
also guarantee the consistency of the whole object. This way, we
increase the overall security of the system.

Resolves #69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Crossover: use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+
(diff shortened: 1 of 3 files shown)


CPerezz (approved):
LGTM. But please, could you merge on a pre-release branch instead of a release one? Also, maybe we should discuss with @ZER0 wether we should already start pushing things here to the `-rc.0` branch so that we can release the RC in the future once Poseidon and the rest is done.

PR #71, opened 2021-04-14
message: use attributes as hash inputs

### LINKED ISSUE #69: crossover: use all attributes as hash inputs

vlopes11:
The message structure follows the same pattern as crossover. It must
implement the same hashing strategies.

Resolves #69


CPerezz (approved):
Looks like it could be simplified as a trait considering that #70 basically did the exact same thing. But you'll know better. 

For what it refers to code, LGTM!

Please run rustfmt before merging!

PR #75, opened 2021-05-03
Release 0.11

### LINKED ISSUE #74: Release phoenix-core v0.11

vlopes11:
Resolves #74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change `Crossover` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
 - Change `Message` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
 - Update `canonical` from `v0.5` to `v0.6` [#72](https://github.com/dusk-network/phoenix-core/issues/72)
+- Change note position to reference [#76](https://github.com/dusk-network/phoenix-core/issues/76)
 
 ## [0.10.0] - 2021-04-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Change `Crossover` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
 - Change `Message` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+- Update `canonical` from `v0.5` to `v0.6` [#72](https://github.com/dusk-network/phoenix-core/issues/72)
 
 ## [0.10.0] - 2021-04-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Change `Crossover` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+- Change `Message` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
 
 ## [0.10.0] - 2021-04-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Change `Crossover` to use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+
 ## [0.10.0] - 2021-04-06
 
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,8 @@ edition = "2018"
 [dependencies]
 rand_core = "0.6"
 dusk-bytes = "0.1"
-dusk-bls12_381 = { version = "0.8.0-rc", default-features = false }
-dusk-jubjub = { version = "0.10.0-rc", default-features = false }
+dusk-bls12_381 = { version = "0.8", default-features = false }
+dusk-jubjub = { version = "0.10", default-features = false }
 dusk-poseidon = { version = "0.21.0-rc", default-features = false }
 dusk-pki = { version = "0.7.0-rc", default-features = false }
 canonical = { version = "0.6", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,18 +5,17 @@ authors = ["zer0 <matteo@dusk.network>", "Victor Lopez <victor@dusk.network"]
 edition = "2018"
 
 [dependencies]
-rand_core = "0.5"
+rand_core = "0.6"
 dusk-bytes = "0.1"
-dusk-bls12_381 = {version = "0.6", default-features = false}
-dusk-jubjub = {version = "0.8", default-features = false}
-dusk-poseidon = {version = "0.20", default-features = false}
-dusk-pki = {version = "0.6", default-features = false}
-canonical = {version = "0.5", optional = true}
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Crossover: use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+- Message: use all attributes as hash inputs [#69](https://github.com/dusk-network/phoenix-core/issues/69)
+- Update `canonical` from `v0.5` to `v0.6` [#72](https://github.com/dusk-network/phoenix-core/issues/72)
+- Change note position to reference [#76](https://github.com/dusk-network/phoenix-core/issues/76)
+
```
ZER0, line 14: It looks like you didn't address my comments in the previous review. This is not a valid format for either the changelog and the commit messages, please change it. This format is reserved to workspaces. Here it looks like `Message` and `Crossover` are members of a phoenix-core workspace.
Also, do not start with `use`.

Those should be rephrase into ```Change `Crossover` to use all attributes as hash inputs```, etc.

Modify the commit messages fa97e741c7a38a3050de929eaf57345c5f75664e and fa97e741c7a38a3050de929eaf57345c5f75664e accordingly.

Cargo.toml
```diff
@@ -1,22 +1,22 @@
...
-dusk-poseidon = {version = "0.20", default-features = false}
-dusk-pki = {version = "0.6", default-features = false}
-canonical = {version = "0.5", optional = true}
-canonical_derive = {version = "0.5", optional = true}
+dusk-bls12_381 = {version="0.8", default-features=false}
+dusk-jubjub = {version="0.10", default-features=false}
+dusk-poseidon = {version="0.21.0-rc", default-features=false}
+dusk-pki = {version="0.7.0-rc", default-features=false}
+canonical = {version="0.6", optional=true}
+canonical_derive = {version="0.6", optional=true}
```
ZER0, line 15: Is there any reason why you remove the space around `=`?

src/crossover.rs
```diff
@@ -68,25 +66,37 @@ impl Serializable<{ 64 + PoseidonCipher::SIZE }> for Crossover {
...
 impl Crossover {
-    /// Returns a hash represented by `H(value_commitment)`
-    pub fn hash(&self) -> BlsScalar {
-        let value_commitment = self.value_commitment().to_hash_inputs();
+    /// Represent the crossover as a sequence of scalars to be used as input for
+    /// sponge hash functions
+    pub fn to_hash_inputs(
+        &self,
+    ) -> [BlsScalar; 3 + PoseidonCipher::cipher_size()] {
+        let mut inputs = [BlsScalar::zero(); 3 + PoseidonCipher::cipher_size()];
```
ZER0, line 74: Where this `3` comes from? You should either add a clear comment or one or more const with a clear name.

src/crossover.rs
```diff
@@ -68,25 +66,37 @@ impl Serializable<{ 64 + PoseidonCipher::SIZE }> for Crossover {
...
+
+        inputs[0..2].copy_from_slice(&self.value_commitment().to_hash_inputs());
```
ZER0, line 76: You can omit `0`.

src/error.rs
```diff
@@ -27,16 +27,21 @@ pub enum Error {
...
-        Error::PoseidonError
+impl From<PoseidonError> for Error {
+    fn from(p: PoseidonError) -> Self {
+        Self::PoseidonError(p)
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Phoenix-Core Error: {:?}", &self)
```
ZER0, line 45: I don't think this should be the way to format our errors, but since we do not have a standard yet we can leave as is for the time being.

src/message.rs
```diff
@@ -61,18 +60,37 @@ impl Message {
         }
     }
 
+    /// Represent the message as a sequence of scalars to be used as input for
+    /// sponge hash functions
+    pub fn to_hash_inputs(
+        &self,
+    ) -> [BlsScalar; 3 + PoseidonCipher::cipher_size()] {
```
ZER0, line 67: Same here about the number `3`.

src/message.rs
```diff
@@ -61,18 +60,37 @@ impl Message {
...
+        let mut inputs = [BlsScalar::zero(); 3 + PoseidonCipher::cipher_size()];
+
+        inputs[0..2].copy_from_slice(&self.value_commitment().to_hash_inputs());
```
ZER0, line 70: Also here you can omit `0`.

src/note.rs
```diff
@@ -237,8 +237,8 @@ impl Note {
     }
 
     /// Return the position of the note on the tree.
-    pub fn pos(&self) -> u64 {
-        self.pos
+    pub fn pos(&self) -> &u64 {
```
ZER0, line 240: Since you modify in message and crossover the methods to be `const` whenever possible, you might do the same for note as well.

Cargo.toml
```diff
@@ -1,22 +1,22 @@
...
 
 [dev-dependencies]
 assert_matches = "1.3"
-rand = "0.7"
+rand = "0.8"
```
ZER0, line 19: We shouldn't have `rand` in dev-dependecies since `rand-core` should suffice. All the core libraries like bls and jubjub now are using `rand-core` and `OsRng` so we should be good to go.
If there is some other libraries used as deps here that still requires `rand` we should probably update them first and `phoenix-core` later.

tests/crossover.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use dusk_jubjub::JubJubScalar;
+use dusk_pki::SecretSpendKey;
+use phoenix_core::{Error, Message, Note};
+
+#[test]
+fn crossover_hash() -> Result<(), Error> {
+    let rng = &mut rand::thread_rng();
+
+    let ssk = SecretSpendKey::random(rng);
```
ZER0, line 17: See my comment about `rand-core` / `rand` in `dev-dependencies`. I believe `dusk-pki` is using `rand-core` so we shouldn't have the need to use `rand` as deps, and we should be able to use `OsRng` directly.

ZER0 (changes requested):
Few nits to address before the release.

CPerezz (approved):
LGTM once @ZER0 's comments are addressed

CHANGELOG.md
vlopes11: Fixed

src/crossover.rs
vlopes11: We have inside this function 3 scalars + whatever comes from the poseidon cipher:

https://github.com/dusk-network/phoenix-core/blob/ba2450df3d0045d3dd7a0ecfab0b5a173cfe1f49/src/crossover.rs#L71-L81

src/crossover.rs
vlopes11: Fixed

Cargo.toml
vlopes11: It was to save space and reduce long features list to a single line.

As agreed, we should leave the space. Fixed

src/message.rs
vlopes11: Fixed

src/note.rs
vlopes11: Fixed

tests/crossover.rs
vlopes11: Fixed

Cargo.toml
vlopes11: Fixed

src/error.rs
vlopes11: This format is the same of PoseidonError. I agree we should set a better standard.

src/crossover.rs
ZER0: As discussed during the stand down, just a comment in this method would suffice about where `3` is coming from (same for `Message`)

vlopes11: Fixed

src/message.rs
vlopes11: Fixed

ZER0 (approved):
LGTM

Issue #78, opened 2021-05-18
Prepare crate for publishing


CPerezz:
Err obtained while doing a dry-run publish:

`warning: manifest has no description, license, license-file, documentation, homepage or repository.`

PR #79, opened 2021-05-18
Add publishing required fields

CPerezz:
Resolves: #78


ZER0 (approved):
LGTM, just rephrase the commit message to have `Resolves: #78` instead of `Resolves: 78`.

PR #81, opened 2021-05-19
Change `rand_core` to not use default features

### LINKED ISSUE #80: `rand_core` shouldn't require default features
phoenix-core is a no-std library. If rand_core isn't required with default-features set to false, then getrandom is pulled. Getrandom won't work in no-std

vlopes11:
The default features of `rand_core` will pull `getrandom` that relies on
`std`.

Since `phoenix-core` is `no-std`, this would lead to major
inconsistencies on the contracts layer.

Resolves #80

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "MPL-2.0"
 exclude = [".github/workflows/ci.yml", ".gitignore"]
 
 [dependencies]
-rand_core = "0.6"
+rand_core = { version = "0.6", default-features = false }
 dusk-bytes = "0.1"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "phoenix-core"
-version = "0.11.0-rc.0"
+version = "0.11.0-rc.1"
```
CPerezz, line 3: Bump on a sepparated commit once this gets merged. This is the procedure we always use.

CPerezz (approved):
LGTM. Just revert the bump and do it in a commit once merged into `release-0.11`

PR #106, opened 2022-03-09
Add `Note::transparent_stealth` function

ureeves:
The function allows the generation of a note by using just a stealth address, a value, and a nonce, in contrast with `Note::transparent` which requires a public spend key. This is required for environments where knowledge of the public spend key is undesirable, but a stealth address is acceptable.

This will be invaluable for the implementation of [this issue](https://github.com/dusk-network/rusk/issues/614).


vlopes11 (approved):
LGTM if CI passes

PR #117, opened 2022-12-06
Rusk transfer contract types 116

OsDusk:
Adopt + adapt transfer contract struct and functions from Rusk repo

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -92,7 +92,7 @@ impl Transaction {
             bytes.extend(val.to_bytes());
         });
 
-        let size = self.nullifiers.len() as u64;
+        let size = self.outputs.len() as u64;
         bytes.extend(size.to_bytes());
         self.outputs.iter().for_each(|val| {
             bytes.extend(val.to_bytes());
@@ -179,11 +179,6 @@ impl Transaction {
             proof,
             call,
         })
-    }
-
-    /// Returns the hash of the transaction.
-    pub fn hash(&self) -> BlsScalar {
-        todo!()
     }
 
     /// Returns the inputs to the transaction.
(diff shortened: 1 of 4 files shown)


src/transaction.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut array = [BlsScalar::default(); STCT_MESSAGE_SIZE];
+    let hash_inputs = crossover.to_hash_inputs();
+    array[..hash_inputs.len()].copy_from_slice(&hash_inputs);
+    array[hash_inputs.len()..].copy_from_slice(&[value.into(), module_id]);
+    array
+}
+
+/// Signature message used for [`Stco`].
+#[must_use]
+pub fn sign_message_stco(
```
miloszm, line 104: as above

src/transaction.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// A call to a contract. The `Vec<u8>` must be an `rkyv`ed representation
+    /// of the data the contract expects, and the `String` the name of the
+    /// function to call.
+    pub call: Option<(ModuleId, String, Vec<u8>)>,
+}
+
+impl Transaction {
+    /// Return input bytes to a hash function for the transaction.
+    #[must_use]
+    pub fn to_hash_input_bytes(&self) -> Vec<u8> {
```
ureeves, line 61: It might be wise to introduce another function that computes the hash input bytes from the decomposed inputs. This is because the `tx_hash` has to be computed before a transaction can be properly created.

```rust
pub fn hash_input_from_components(
        nullifiers: &[BlsScalar],
        outputs: &[Note],
        anchor: &BlsScalar,
        fee: &Fee,
        crossover: &Option<Crossover>,
        call: &Option<(ModuleId, String, Vec<u8>)>,
    ) {
// ...
}
```
The current function would then call this one.

ureeves (changes requested):
One thing I think should happen before merging

src/transaction.rs
ureeves: Also, there's the issue of serialization of the transaction. We should probably have a `to_var_bytes` function that serializes a transaction, and a `from_slice` or `from_bytes` that deserializes it.

PR #120, opened 2023-01-19
Add types to interact with the genesis contracts

### LINKED ISSUE #119: Add functions and data structures for interacting with the genesis contracts
#### Summary
During the [`rusk` migration to `piecrust`](https://github.com/dusk-network/rusk/issues/760), it has become clear that it is necessary to have a place for structures used as arguments for the genesis smart contracts. Since we've already included the concept of a `Transaction` in this crate, it also makes sense to include other components of the transaction model, such as sending and withdrawing value to and from a contract, or staking and unstaking.

Included in the [`*-types` crates](https://github.com/dusk-network/rusk/blob/b5c4cca4f5c15f614b6f169904f0d772e08e93ff/contracts/transfer-types/src/lib.rs#L168) are the structures that should be ported to this crate. The goal is to eliminate the `*-types` crates.

#### Additional context
If a contract writer wishes to interact with the genesis contracts - where Dusk's transaction model is supported - it makes sense to include this crate to do so. As such, this crate should support this use-case.

ureeves:
Resolves #119

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -6,6 +6,12 @@
 
 //! Verifier data for the transfer circuits
 
+mod stake;
+pub use stake::*;
+
+mod transfer;
+pub use transfer::*;
+
 use alloc::string::String;
 use alloc::vec::Vec;
 
@@ -14,12 +20,11 @@ use rkyv::{Archive, Deserialize, Serialize};
 
 use dusk_bls12_381::BlsScalar;
 use dusk_bytes::{DeserializableSlice, Error as BytesError, Serializable};
-use dusk_poseidon::cipher::PoseidonCipher;
 
-use crate::{Crossover, Fee, Message, Note};
+use crate::{Crossover, Fee, Note};
 
-const STCO_MESSAGE_SIZE: usize = 7 + 2 * PoseidonCipher::cipher_size();
-const STCT_MESSAGE_SIZE: usize = 5 + PoseidonCipher::cipher_size();
+/// Type alias for the ID of a module.
+pub type ModuleId = [u8; 32];
 
 /// A phoenix transaction.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


src/transaction.rs
```diff
@@ -14,12 +20,11 @@ use rkyv::{Archive, Deserialize, Serialize};
...
 use dusk_bytes::{DeserializableSlice, Error as BytesError, Serializable};
-use dusk_poseidon::cipher::PoseidonCipher;
 
-use crate::{Crossover, Fee, Message, Note};
+use crate::{Crossover, Fee, Note};
 
-const STCO_MESSAGE_SIZE: usize = 7 + 2 * PoseidonCipher::cipher_size();
-const STCT_MESSAGE_SIZE: usize = 5 + PoseidonCipher::cipher_size();
+/// Type alias for the ID of a module.
+pub type ModuleId = [u8; 32];
```
moCello, line 27: Although this is clearly an improvement to the previous version where we don't add a type alias and only use `BlsScalars` as an ID, wouldn't it make sense to use the piecrust `ModuleId` here even though it would include an extra dependency? 
It doesn't feel 'right' to define the same thing in two different crates.

moCello (approved):
LGTM, just one question but no blocker

src/transaction.rs
ureeves: There's another example of this in this very crate already, and that is a `dusk_plonk::Proof`. The `proof` field of a `Transaction` really _should be_ a `Proof`, however, in the interest of reducing dependencies `dusk_plonk` hasn't been included in the manifest and the field is a `Vec<u8>` instead.
In fact, this example is even worse, since to convert back and forth between `Proof` and `Vec<u8>` is not `const`, while converting between `[u8; 32]` and `piecrust_uplink::ModuleId` is.
I suggest an issue be created to discuss how to deal with this.

ureeves: Created issue https://github.com/dusk-network/phoenix-core/issues/121 to deal with this.

PR #124, opened 2023-01-30
Remove `enc`, `R` and `nonce` from note hash

### LINKED ISSUE #123: Remove `enc`, `nonce` and `R` from the hash of a note
Currently, when we compute the hash of a note (in the circuit and in the Merkle tree of notes), we include the parameters named `enc`, `nonce`, and `R` (where R is a part of the stealth address). We figured that these three elements are not needed in this computation. Hence, they can be removed from the circuit. This change will affect the circuit and the storage of notes in the Merkle tree of notes.

xevisalle:


ureeves (changes requested):
Please add a CHANGELOG entry

ureeves (approved):
LGTM

PR #129, opened 2023-05-27
Update CI

### LINKED ISSUE #128: Implement Standardized CI Workflow
#### Summary
We've recently introduced a set of reusable workflow files in our [.github](https://github.com/dusk-network/.github/tree/main/.github/workflows#readme) repository to standardize our CI/CD process for Rust projects. These new workflows use up-to-date, maintained actions and also introduce caching, leading to significant performance improvements in subsequent runs. This repo is a prime candidate to benefit from this new setup.

To implement this, we'll need to replace our current workflow process with the new workflow format. This involves removing the current jobs and replacing them with calls to the reusable workflows. We need to carefully test and check that the new workflows actually execute as expected.

#### Additional context
Due to Rust being available by default on the target OS, a `rust-toolchain` file is enough to specify what Rust version/target is required for a given workflow.

HDauven:
Resolves #128


.github/workflows/dusk_ci.yml
```diff
@@ -1,72 +1,29 @@
...
-      - run: rustup component add rustfmt
-      - uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+  test_rkyv:
+    name: tests rkyv
+    uses: dusk-network/.github/.github/workflows/run-tests.yml@main
+    with:
+      test_flags: --features rkyv-impl
```
ureeves, line 29: TBH, I find having special tests for `rkyv` a little strange. I think I understand why, but could it be possible to do this differently?

ureeves (approved):
LGTM. I do find `rkyv` specific stuff a little strange though, could you talk more about it?

.github/workflows/dusk_ci.yml
HDauven: @ureeves the `run-tests` job, by default, runs `cargo test --release` if no test flags are passed. 

If you want to also test a specific feature, you need to pass it like I did here.

You could get rid of either the `test_std` job or `rkyv-impl` if you're fine with running the test and not using it as a sort of "integration test" to run both with or without the feature enabled.

ureeves: Ah understood. The workflow is essentially a frontend to `cargo`, and we can pass flags as we please.

HDauven: Yup! For more details on how you can use it, see the README here: https://github.com/dusk-network/.github/tree/main/.github/workflows#readme

This is the underlying workflow being called for `cargo test`s: https://github.com/dusk-network/.github/blob/main/.github/workflows/run-tests.yml

PR #133, opened 2023-11-02
 Add Phoenix protocol's related keys from dusk_pki

Neotamandua:
`PublicSpendKey` is now `PublicKey`
`SecretSpendKey` is now `SecretKey`

The next step is to refactor the code, to align with the desired internal representation of `PublicKey` & `SecretKey` (will be separate PR).

### DIFF
--- a/src/keys/public.rs
+++ b/src/keys/public.rs
@@ -1,0 +1,105 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use crate::{permutation, SecretKey, StealthAddress};
+
+use dusk_jubjub::{JubJubAffine, JubJubExtended, JubJubScalar};
+
+#[cfg(feature = "rkyv-impl")]
+use rkyv::{Archive, Deserialize, Serialize};
+
+use dusk_bytes::{DeserializableSlice, Error, HexDebug, Serializable};
+use dusk_jubjub::GENERATOR_EXTENDED;
+use subtle::{Choice, ConstantTimeEq};
+
+/// Public pair of `a·G` and `b·G` defining a [`PublicKey`]
+#[derive(HexDebug, Clone, Copy)]
+#[cfg_attr(
+    feature = "rkyv-impl",
+    derive(Archive, Serialize, Deserialize),
+    archive_attr(derive(bytecheck::CheckBytes))
+)]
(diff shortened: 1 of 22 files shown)


src/permutation.rs
```diff
@@ -0,0 +1,13 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use dusk_jubjub::{JubJubExtended, JubJubScalar};
+use dusk_poseidon::sponge::truncated;
+
+/// Hashes a JubJub's ExtendedPoint into a JubJub's Scalar
```
moCello: maybe add that it's using the poseidon hash function

src/keys/secret.rs
```diff
@@ -0,0 +1,112 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use super::public::PublicKey;
+use super::stealth::StealthAddress;
```
moCello: can't you use `crate::...` here?

src/keys/secret.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::{permutation, ViewKey};
```
moCello: and merge with this :)

src/keys/public.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn gen_stealth_address(&self, r: &JubJubScalar) -> StealthAddress {
+        let G = GENERATOR_EXTENDED;
+        let R = G * r;
+
+        let rA = self.A * r;
+        let rA = permutation::hash(&rA);
+        let rA = G * rA;
+
+        let pk_r = rA + self.B;
+        let pk_r = NotePublicKey::from(pk_r);
```
moCello: I'm not sure if I still like this even though I suggested it earlier.

Maybe this actually also works and looks a bit cleaner
```suggestion
        let pk_r: NotePublicKey = pk_r.into();
```

src/keys/secret.rs
moCello: I now see that this is also happening in other places

moCello (approved):
LGTM, two tiny nits but no blocker

src/keys/secret.rs
Neotamandua: thank you, will check it out

PR #138, opened 2023-12-07
Use `From` trait to generate keys

### LINKED ISSUE #136: Implement `From<SecretKey>` for `PublicKey` and `ViewKey`
#### Summary
The generation of public keys should be the same across our stack.
In our signature scheme implementations we always derive a public key from a secret key using the `From` trait. This needs to be implemented for the `PublicKey` and `ViewKey` in this library as well.

Additionally, the generation of the `StealthAddress` should also align with this as much as possible.

moCello:
Resolves #136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Removed
+
+- Remove `HexDebug` trait for keys [#136]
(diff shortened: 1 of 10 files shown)


herr-seppia (approved):
LGTM

Why did you remove the HexDebug? I liked it

moCello:
Since we merged the bls and jubjub curve with the upstream code, we don't derive `HexDebug` for scalars anymore, instead we use their `Debug` implementation which should give the same result as our previous `HexDebug`.

ureeves (approved):

PR #151, opened 2024-04-05
Add AES cipher module

### LINKED ISSUE #152: Implement and use the AES cipher instead of PoseidonCipher
# Epic

## Summary
Atm we were using PoseidonCipher to encrypt the value / blinder in the notes. While it was a requirement for in-circuit encryption, this is not the case anymore as we got rid of the send to contract obfuscated approach. We can now use a faster solution instead, like AES. In particular, we've chosen AES-GCM (https://crates.io/crates/aes-gcm), given that: 

- it's efficient
- it's approved by NIST
- the crate has been audited
- the crate is widely used

xevisalle:
Resolves: https://github.com/dusk-network/phoenix-core/issues/152

**Summary**

- We create the `cipher` module, which uses AES-GCM for encryption.
- We use the `cipher` module throughout the code, removing `PoseidonCipher`.
- We remove the `Message` module, only meant to be used along with the `send to / withdraw from contract obfuscated`, whose feature will be removed too.


herr-seppia (commented):
Some questions:
- Are those unwrap/expect inside encrypt/decrypt intentional? Wouldn't a Result fit better? 
- Any reason to remove the "deterministic" creation of the note?

src/note.rs
```diff
@@ -68,10 +76,9 @@ impl TryFrom<i32> for NoteType {
 pub struct Note {
     pub(crate) note_type: NoteType,
     pub(crate) value_commitment: JubJubExtended,
-    pub(crate) nonce: BlsScalar,
     pub(crate) stealth_address: StealthAddress,
     pub(crate) pos: u64,
-    pub(crate) encrypted_data: PoseidonCipher,
+    pub(crate) encryption: [u8; ENCRYPTION_SIZE],
```
herr-seppia, line 82: Shouldn't this be called just "data"?

src/note.rs
```diff
@@ -107,39 +144,43 @@ impl Note {
...
         let value_commitment = JubJubScalar::from(value);
         let value_commitment = (GENERATOR_EXTENDED * value_commitment)
             + (GENERATOR_NUMS_EXTENDED * TRANSPARENT_BLINDER);
 
         let pos = u64::MAX;
 
-        let zero = TRANSPARENT_BLINDER.into();
-        let mut encrypted_data = [zero; PoseidonCipher::cipher_size()];
+        let mut encryption = value.to_bytes().to_vec();
+        encryption.append(&mut [0; ENCRYPTION_SIZE - u64::SIZE].to_vec());
```
herr-seppia: The `- u64::SIZE` is very weird imo

src/cipher.rs
```diff
@@ -0,0 +1,89 @@
...
+    let cipher = Aes256Gcm::new(key);
+    let nonce = Aes256Gcm::generate_nonce(rng);
+    let mut ciphertext = cipher
+        .encrypt(&nonce, plaintext.as_ref())
+        .expect("Encryption failed.");
+
+    let mut encryption = nonce.to_vec();
+    encryption.append(&mut ciphertext);
+
+    encryption.try_into().unwrap()
```
moCello, line 42: If you use `unwrap` it's good to add a comment that explains why this can not panic.
Even better: use `expect` and add the reason in the message (e.g. "this can not fail because ..." or "this will always pass because ...")

src/cipher.rs
```diff
@@ -0,0 +1,89 @@
...
+use aes_gcm::{
+    aead::{Aead, AeadCore, KeyInit},
+    Aes256Gcm, Key,
+};
+
+const NONCE_SIZE: usize = 12;
+
+/// Size of the extra encryption data required by the
+/// cipher: the nonce (12 bytes) and the tag (16 bytes)
+pub const EXTRA_DATA_SIZE: usize = NONCE_SIZE + 16;
```
herr-seppia: Should 16 be declared as `TAG_SIZE`?

src/cipher.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let nonce = &encryption[..NONCE_SIZE];
+    let ciphertext = &encryption[NONCE_SIZE..];
+
+    let cipher = Aes256Gcm::new(key);
+    cipher
+        .decrypt(nonce.into(), ciphertext.as_ref())
+        .expect("Decryption failed.")
+        .try_into()
+        .unwrap()
```
moCello, line 64: same

src/cipher.rs
```diff
@@ -0,0 +1,89 @@
...
+    let key = secret_key.to_bytes();
+    let key = Key::<Aes256Gcm>::from_slice(&key);
+
+    let nonce = &encryption[..NONCE_SIZE];
+    let ciphertext = &encryption[NONCE_SIZE..];
+
+    let cipher = Aes256Gcm::new(key);
+    cipher
+        .decrypt(nonce.into(), ciphertext.as_ref())
+        .expect("Decryption failed.")
```
moCello, line 62: What are the reasons that the decryption can fail? If they can happen with normal use, it might be a better idea to return a `Result<[u8; PLAINTEXT_SIZE], Error> here so that the user can decide on how to recover from the error instead of the entire program simply panicing.

src/note.rs
xevisalle: In the specs we use `enc` to refer to this field. I think it's important to know that this contains a ciphertext. We could call it `encryption_data` but we can make it shorter. The former `encrypted_data` wouldn't be correct now as this value contains a nonce which is not encrypted.

src/crossover.rs
```diff
@@ -6,37 +6,44 @@
...
-use dusk_poseidon::cipher::PoseidonCipher;
-use dusk_poseidon::sponge;
 
 #[cfg(feature = "rkyv-impl")]
 use rkyv::{Archive, Deserialize, Serialize};
 
+use crate::note::ENCRYPTION_SIZE;
+
 /// Crossover structure containing obfuscated encrypted data
-#[derive(Clone, Copy, Debug, Default)]
```
moCello, line 19: can the `Default` trait not be derived anymore?

src/note.rs
xevisalle: we need to append zeros here... up to the same size of the obfuscated notes. The `- u64::SIZE` refers to the value already in there. What do you propose? Maybe create another constant at the top of the file doing this? Or putting directly the value?

src/cipher.rs
xevisalle: I don't think so. The tag size is something internal of the AES crate. I explained it in the comment so that everyone can understand what the output is, but I think we will never need to "touch" that value.

src/cipher.rs
```diff
@@ -0,0 +1,89 @@
...
+const NONCE_SIZE: usize = 12;
+
+/// Size of the extra encryption data required by the
+/// cipher: the nonce (12 bytes) and the tag (16 bytes)
+pub const EXTRA_DATA_SIZE: usize = NONCE_SIZE + 16;
+
+/// Encrypts a plaintext given a shared DH secret key, returning a vector
+/// containing a nonce and the ciphertext (which includes the tag)
+pub fn encrypt<R: RngCore + CryptoRng, const ENCRYPTION_SIZE: usize>(
+    secret_key: &JubJubAffine,
```
moCello: Can we call it `shared_secret`? I always get confused when I read `secret_key` and then it's a jubjub-point

src/note.rs
```diff
@@ -13,7 +13,9 @@ use dusk_jubjub::{
     dhke, JubJubAffine, JubJubExtended, JubJubScalar, GENERATOR_EXTENDED,
     GENERATOR_NUMS_EXTENDED,
 };
-use dusk_poseidon::cipher::PoseidonCipher;
+
+use crate::cipher::{decrypt, encrypt, EXTRA_DATA_SIZE};
+
 use dusk_poseidon::sponge::hash;
```
moCello, line 19: not directly linked to this PR but are we still using the poseidon hash here?
Cause if not I think we can remove the `dusk-poseidon` dependency entirely!

xevisalle:
> Some questions:
> 
> * Are those unwrap/expect inside encrypt/decrypt intentional? Wouldn't a Result fit better?
> * Any reason to remove the "deterministic" creation of the note?

I will think about the errors... probably you are right.

About the deterministic: the `cipher` module needs randomness to create a nonce. I know that before this was a parameter for the encryption, but as its generation is now very linked to the AES crate we use, I think is better to compute everything into the module. So, this function is no longer "deterministic". However, now the randomness `r` has to be provided when creating a `new` note, so I think that the same functionality we had before is still there.

src/note.rs
```diff
@@ -87,14 +94,44 @@ impl Note {
     pub fn new<R: RngCore + CryptoRng>(
         rng: &mut R,
         note_type: NoteType,
+        r: &JubJubScalar,
```
moCello: Can you comment what `r` is in this context?

src/note.rs
```diff
@@ -285,19 +275,14 @@ impl Note {
...
-
     /// Return the value commitment `H(value, blinding_factor)`
     pub const fn value_commitment(&self) -> &JubJubExtended {
         &self.value_commitment
     }
 
     /// Returns the cipher of the encrypted data
-    pub const fn cipher(&self) -> &[BlsScalar; PoseidonCipher::cipher_size()] {
-        self.encrypted_data.cipher()
+    pub const fn cipher(&self) -> &[u8; ENCRYPTION_SIZE] {
```
moCello: can we call it `encryption`?

tests/note_test.rs
```diff
@@ -175,8 +172,9 @@ fn note_keys_consistency() {
     assert!(vk.owns(&note));
 }
 
+#[should_panic]
 #[test]
-fn fee_and_crossover_generation() -> Result<(), Error> {
```
moCello, line 179: any specific reason you removed the `Result` return? If you leave it there you can use the `?` a few lines down instead of unwrapping

moCello (commented):
Great work!
A few little comments and generally please try to get rid of the `unwrap`. Better return a `Result` and convert the `aes-gcm` error type into an `phoenix-core` error type so you can use the `?` operator. (deleted all my previous separate comments on this to declutter the review)
And another question, is it possible to name the `cipher` module `encryption`? not sure what is the convention here but to me `encryption` makes more sense

src/note.rs
xevisalle: Of course. Good catch

src/cipher.rs
xevisalle: Maybe... or maybe it could require the bytes instead of a point. Mmm..

xevisalle:
> Great work!
> 
> A few little comments and generally please try to get rid of the `unwrap`. Better return a `Result` and convert the `aes-gcm` error type into an `phoenix-core` error type so you can use the `?` operator. (deleted all my previous separate comments on this to declutter the review)
> 
> And another question, is it possible to name the `cipher` module `encryption`? not sure what is the convention here but to me `encryption` makes more sense

Good, I'll create the errors and use the ?. 

Cipher is the algorithm. Encryption is commonly used for the same purpose but AFAIK is a bit more generic. I used it in our context to refer to the pair nonce + ciphertext. I don't have a strong opinion but `encryption = encryption::encrypt()` doesn't look that nice 😔. In any case, as you prefer!

src/cipher.rs
moCello: maybe just call it `secret` then? although in the end, isn't it always going to be a shared secret (a point from the dhke)? Why else would you want to encrypt something if not for someone else to decrypt?

moCello:
Personally I have a preference for `encryption::encrypt` and in lib.rs have a `pub use encryption::{encrypt, decrypt};` re-export.

src/crossover.rs
xevisalle: I had to implement it myself because the new encryption attribute doesn't implements it

src/cipher.rs
xevisalle: ok let's calll it shared_secret or shared_key then!

src/note.rs
xevisalle: still needed to compute the note hash!

src/error.rs
```diff
@@ -25,8 +29,8 @@ pub enum Error {
     InvalidCrossoverConversion,
     /// Invalid Fee for conversion
     InvalidFeeConversion,
-    /// Failure to decrypt
-    Decryption,
+    /// Failure to encrypt / decrypt
+    Encryption,
```
moCello: Would it make sense to call it `InvalidEncryption` or `IncorrectEncryption`? It's a bit more descriptive

src/encryption.rs
```diff
@@ -0,0 +1,90 @@
...
+    let key = Key::<Aes256Gcm>::from_slice(&key);
+
+    let cipher = Aes256Gcm::new(key);
+    let nonce = Aes256Gcm::generate_nonce(rng);
+    let mut ciphertext = cipher.encrypt(&nonce, plaintext.as_ref())?;
+
+    let mut encryption = nonce.to_vec();
+    encryption.append(&mut ciphertext);
+
+    Ok(encryption.try_into()?)
```
moCello: I would try this:
```suggestion
    let mut encryption = [0u8; ENCRYPTION_SIZE];
    encryption[..NONCE_SIZE].copy_from_slice(&nonce);
    encryption[NONCE_SIZE..].copy_from_slice(&ciphertext[..]);

    Ok(encryption)
```

src/encryption.rs
```diff
@@ -0,0 +1,90 @@
...
+    let key = Key::<Aes256Gcm>::from_slice(&key);
+
+    let nonce = &encryption[..NONCE_SIZE];
+    let ciphertext = &encryption[NONCE_SIZE..];
+
+    let cipher = Aes256Gcm::new(key);
+
+    Ok(cipher
+        .decrypt(nonce.into(), ciphertext.as_ref())?
+        .try_into()?)
```
moCello: And here this:
```suggestion
    let mut plain_text = [0u8; PLAINTEXT_SIZE];
    let decryption = cipher.decrypt(nonce.into(), ciphertext)?;

    plain_text.copy_from_slice(&decryption[..]);

    Ok(plain_text)
```

src/error.rs
```diff
@@ -45,6 +49,20 @@ impl fmt::Display for Error {
     }
 }
 
+impl From<alloc::vec::Vec<u8>> for Error {
+    fn from(_vec: Vec<u8>) -> Self {
+        Error::Encryption
+    }
+}
```
moCello: With the changes in `encrypt` and `decrypt` you will not need this conversion.

src/lib.rs
```diff
@@ -21,14 +21,15 @@ pub mod crossover;
...
-/// Message representation
-pub mod message;
 /// Transparent and Obfuscated Notes
 pub mod note;
 
 /// Phoenix Core Keys & Addresses
 mod keys;
 
+/// Encryption and decryption methods
+pub mod encryption;
```
moCello:
```suggestion
mod encryption;
```
I would suggest to not make this a public module but instead re-export the two functions `encrypt` and `decrypt` in lib.rs:
```
pub use encryption::{encrypt, decrypt};
```
This way we avoid the `encryption::encrypt` path and it's generally easier to see which types are exposed and which are not when the modules are private in general.

src/encryption.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[cfg(test)]
```
moCello: Please add this as an integration test, not a unit test.

src/note.rs
```diff
@@ -107,39 +145,43 @@ impl Note {
...
 
-        let zero = TRANSPARENT_BLINDER.into();
-        let mut encrypted_data = [zero; PoseidonCipher::cipher_size()];
+        let mut encryption = value.to_bytes().to_vec();
+        encryption.append(&mut [0; ENCRYPTION_SIZE - u64::SIZE].to_vec());
 
-        encrypted_data[0] = BlsScalar::from(value);
-
-        let encrypted_data = PoseidonCipher::new(encrypted_data);
+        let encryption = encryption.try_into().unwrap();
```
moCello: to me it seems weird that you are creating an array of zeroes that you convert to a vector and then back into an array again.
Can you try to create a
```rust
let mut encryption = [0u8; ENCRYPTION_SIZE];
```
array first and then copy the bytes into it as needed?

src/note.rs
```diff
@@ -155,57 +197,8 @@ impl Note {
         value: u64,
         blinding_factor: JubJubScalar,
     ) -> Self {
-        Self::new(rng, NoteType::Obfuscated, psk, value, blinding_factor)
-    }
-
-    /// Create a new phoenix output note without inner randomness
-    pub fn deterministic(
```
moCello, line 162: You sure we don't need this anymore?

src/note.rs
```diff
@@ -342,7 +328,7 @@ impl Ownable for Note {
     }
 }
 
-impl Serializable<{ 137 + PoseidonCipher::SIZE }> for Note {
+impl Serializable<{ 105 + ENCRYPTION_SIZE }> for Note {
```
moCello, line 325: Please add a comment explaining the 105

src/crossover.rs
```diff
@@ -56,57 +62,24 @@ impl Serializable<{ 64 + PoseidonCipher::SIZE }> for Crossover {
     fn from_bytes(bytes: &[u8; Self::SIZE]) -> Result<Self, Self::Error> {
         let value_commitment =
             JubJubExtended::from(JubJubAffine::from_slice(&bytes[..32])?);
-        let nonce = BlsScalar::from_slice(&bytes[32..])?;
 
-        let encrypted_data = PoseidonCipher::from_slice(&bytes[64..])?;
+        let encryption = bytes[32..].try_into().unwrap();
```
moCello: avoid calling `try_into().unwrap()`.

src/note.rs
```diff
@@ -370,21 +355,19 @@ impl Serializable<{ 137 + PoseidonCipher::SIZE }> for Note {
...
-        let nonce = BlsScalar::from_slice(&bytes[33..65])?;
-        let stealth_address = StealthAddress::from_slice(&bytes[65..129])?;
+        let stealth_address = StealthAddress::from_slice(&bytes[33..97])?;
 
-        one_u64.copy_from_slice(&bytes[129..137]);
+        one_u64.copy_from_slice(&bytes[97..105]);
         let pos = u64::from_le_bytes(one_u64);
 
-        let encrypted_data = PoseidonCipher::from_slice(&bytes[137..])?;
+        let encryption = bytes[105..].try_into().unwrap();
```
moCello: avoid calling `try_into().unwrap()`

src/note.rs
```diff
@@ -87,14 +94,45 @@ impl Note {
...
+
+        // Output notes have undefined position, equals to u64's MAX value
+        let pos = u64::MAX;
 
-        Self::deterministic(note_type, &r, nonce, psk, value, blinding_factor)
+        let encryption = match note_type {
+            NoteType::Transparent => {
+                let mut value = value.to_bytes().to_vec();
+                value.append(&mut [0; ENCRYPTION_SIZE - u64::SIZE].to_vec());
+                value.try_into().expect("Converted correctly.")
```
moCello: here too

src/convert.rs
```diff
@@ -79,28 +75,22 @@ impl From<Remainder> for Note {
...
-            let zero = TRANSPARENT_BLINDER.into();
-            let mut encrypted_data = [zero; PoseidonCipher::cipher_size()];
+        let mut encryption = value.to_bytes().to_vec();
+        encryption.append(&mut [0; ENCRYPTION_SIZE - u64::SIZE].to_vec());
 
-            encrypted_data[0] = BlsScalar::from(value);
-
-            PoseidonCipher::new(encrypted_data)
-        };
+        let encryption = encryption.try_into().unwrap();
```
moCello: here too

moCello (commented):
Nice!
A general remark:
I would really try to avoid `try_into`. It makes you need to `unwrap` where it's not really needed.
If you generate an array that has the correct size and then copy the bytes you need, you have more control on what's happening and avoid `unwrap`.

src/error.rs
```diff
@@ -4,9 +4,12 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use aes_gcm::Error as AesError;
 use core::fmt;
 use dusk_bytes::{BadLength, Error as DuskBytesError, InvalidChar};
 
+extern crate alloc;
```
moCello: I don't think you need this anymore.

moCello (approved):
Very nice!
one last left-over from the last changes but otherwise approved!

src/note.rs
```diff
@@ -93,14 +93,14 @@ impl Eq for Note {}
 impl Note {
     /// Creates a new phoenix output note
     pub fn new<R: RngCore + CryptoRng>(
-        rng: &mut R,
+        mut rng: &mut R,
```
moCello: remove `mut`

src/note.rs
```diff
@@ -93,14 +93,14 @@ impl Eq for Note {}
...
         note_type: NoteType,
-        r: &JubJubScalar,
         psk: &PublicKey,
         value: u64,
         blinding_factor: JubJubScalar,
     ) -> Self {
-        let stealth_address = psk.gen_stealth_address(r);
+        let r = JubJubScalar::random(&mut rng);
```
moCello: the trick is to do `&mut *rng`

PR #157, opened 2024-04-19
Rename `sk_r` and `pk_r` to `note_..`

### LINKED ISSUE #156: Rename `SecretKey::sk_r` to `generate_note_sk` and `StealthAddress::pk_r` to `note_pk`
#### Summary
Rename `SecretKey::sk_r` to `generate_note_sk` and `StealthAddress::pk_r` to `note_pk` to make the methods more descriptive.

moCello:
Resolves #156

### DIFF
--- a/src/keys/public.rs
+++ b/src/keys/public.rs
@@ -44,7 +44,7 @@ impl PublicKey {
         &self.B
     }
 
-    /// Generates new `PKr = H(A · r) · G + B` from a given `r`
+    /// Generates new `note_pk = H(A · r) · G + B` from a given `r`
     pub fn gen_stealth_address(&self, r: &JubJubScalar) -> StealthAddress {
         let G = GENERATOR_EXTENDED;
         let R = G * r;
@@ -53,10 +53,10 @@ impl PublicKey {
         let rA = hash(&rA);
         let rA = G * rA;
 
-        let pk_r = rA + self.B;
-        let pk_r = pk_r.into();
+        let note_pk = rA + self.B;
+        let note_pk = note_pk.into();
 
-        StealthAddress { R, pk_r }
+        StealthAddress { R, note_pk }
     }
 }
 
--- a/src/keys/stealth.rs
+++ b/src/keys/stealth.rs
@@ -16,7 +16,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 
 /// To obfuscate the identity of the participants, we utilizes a Stealth Address
 /// system.
-/// A `StealthAddress` is composed by a one-time public key (`pk_r`, the actual
+/// A `StealthAddress` is composed by a one-time note-public-key (the actual
 /// address) and a random point `R`.
 #[derive(Default, Debug, Clone, Copy)]
 #[cfg_attr(
@@ -26,7 +26,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 )]
 pub struct StealthAddress {
     pub(crate) R: JubJubExtended,
-    pub(crate) pk_r: NotePublicKey,
+    pub(crate) note_pk: NotePublicKey,
 }
 
 /// The trait `Ownable` is required by any type that wants to prove its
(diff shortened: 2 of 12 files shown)


src/keys/stealth.rs
```diff
@@ -48,30 +48,30 @@ impl StealthAddress {
...
-    pub const fn pk_r(&self) -> &NotePublicKey {
-        &self.pk_r
+    /// Gets the `note_pk`
+    pub const fn note_pk(&self) -> &NotePublicKey {
+        &self.note_pk
     }
 
-    /// Gets the underline `JubJubExtended` point of `pk_r`
+    /// Gets the underline `JubJubExtended` point of `note_pk`
     pub fn address(&self) -> &JubJubExtended {
```
xevisalle: as this is returning the `note_pk`, I'd call this function like this. In our context the address is the static public key. Otherwise if we refer to the stealth address, we must consider the tuple `(note_pk, R)`

xevisalle (changes requested):
Just the comment about the address.

Plus, as a reflexion, I'd prefer the variables to be called `nsk` and `npk`, but I understand that calling em `note_sk` and `note_pk` is much more readable, so it's fine to me!

src/keys/public.rs
HDauven, line 1: On line 47, docs could be turned into: 
```rust
/// Generates new `note_pk = H(A · r) · G + B` from a given `r`
```

src/keys/view.rs
HDauven, line 1: You could maybe turn line 67 into this:
```rust
/// Checks `note_pk = H(R · a) · G + B`
```

HDauven (approved):
LGTM

Added two places where the docs might be updated. Will approve and leave it up to you whether to change it or not.

src/keys/stealth.rs
moCello: I'm also not entirely happy with the `address` function.

We already have a method called `note_pk` (formerly `pk_r`) of the `StealthAddress` that returns a `NotePublicKey` (which is just a wrapped `JubJubExtended`) so I cannot use the same function name here again.

If we want to get the internal point in `NotePublicKey` we currently have two options:
```rust
// without `address` method
let jubjub_point = stealth_address.note_pk().as_ref();

// with `address` method
let jubjub_point = stealth_address.address();
```
I think the second one is nicer but of course only if the concept of an 'address' somewhat makes sense.

Any thoughts?

moCello:
> Plus, as a reflexion, I'd prefer the variables to be called `nsk` and `npk`, but I understand that calling em `note_sk` and `note_pk` is much more readable, so it's fine to me!

Yes, that's also an option. I'll do it like this in the changes and the stack bump because if I follow the convention `npk` for `NotePublicKey`, then I'll need to name the `StakeSecretKey` in the stake contract `ssk` which will be very confusing with the formerly known `SecretSpendKey`. To avoid such confusion I'll stick to the longer names for now.
In the long run I think we'll see both options in our crates.

src/keys/stealth.rs
ureeves: Adding my two cents here. I think having a `StealthAddress::address` function is a bit... redundant to say the least, and confusing in the worst case. That said, I also understand the `stealth_address.note_pk().as_ref()` is unsatisfactory.

Maybe we should improve the access to the underlying `JubJubScalar` in `NotePublicKey` in a future PR?

ureeves (approved):
Very nice. Approved. Added my two cents to the `address` debate

src/keys/stealth.rs
moCello: I agree and removed the `address` method. I also agree that we need to address the `as_ref()` API, this needs to be done in `jubjub-schnorr` though.

moCello: https://github.com/dusk-network/jubjub-schnorr/issues/13

xevisalle (approved):
LGTM

PR #158, opened 2024-04-23
Implement `Zeroize` for `SecretKey`

### LINKED ISSUE #155: Secret values not erased
#### Summary
Add a way to securely erase sensitive data from memory when types go out of scope.

Add `Zeroize` trait, remove `Copy` trait on all structs carrying sensitive data.

#### Additional context
See [bls12_381-bls #5](https://github.com/dusk-network/bls12_381-bls/issues/5)

moCello:
Resolves #155

### DIFF
--- a/src/keys/secret.rs
+++ b/src/keys/secret.rs
@@ -8,6 +8,7 @@ use crate::{keys::hash, Ownable};
 use dusk_jubjub::JubJubScalar;
 use ff::Field;
 use jubjub_schnorr::SecretKey as NoteSecretKey;
+use zeroize::Zeroize;
 
 #[cfg(feature = "rkyv-impl")]
 use rkyv::{Archive, Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


src/keys/view.rs
```diff
@@ -77,12 +77,6 @@ impl ViewKey {
     }
 }
 
-impl From<SecretKey> for ViewKey {
-    fn from(sk: SecretKey) -> Self {
-        Self::from(&sk)
-    }
-}
-
```
ureeves, line 85: That's quite the catch on the by-copy. Nice

ureeves (approved):
Seeing all the `PublicKey::from(&sk)` make me feel there should be a `SecretKey::public` function. That said, not the purpose here. LGTM

HDauven (approved):
LGTM

src/keys/view.rs
moCello: indeed! Will have some ripple effects downstream, but good ones!

src/keys/secret.rs
```diff
@@ -17,7 +18,30 @@ use rand_core::{CryptoRng, RngCore};
 use subtle::{Choice, ConstantTimeEq};
 
 /// Secret pair of `a` and `b` defining a [`SecretKey`]
-#[derive(Clone, Copy, Eq, Debug)]
+///
+/// ## Safety
+///
+/// To ensure that no secret information lingers in memory after the variable
+/// goes out of scope, we advice calling `zeroize` before the variable goes out
+/// of scope.
```
Daksh14, line 26: Would implementing `Drop` and calling `zeroize` there solve this problem?

moCello: Sadly not. I investigated the problem in our bls signature library: https://github.com/dusk-network/bls12_381-bls/pull/6
long story short: rust still often copies values internally, so the zeroize on drop often doesn't result in actually erasing the desired values.

PR #161, opened 2024-04-29
Add ElGamal cipher

### LINKED ISSUE #162: Implement ElGamal cipher
#### Summary
We need to implement the ElGamal cipher to create recipient proofs. As such, we need this asymmetric cipher to be computed in-circuit, thus a gadget will be needed.

xevisalle:
Resolves: https://github.com/dusk-network/phoenix-core/issues/162


tests/encryption.rs
```diff
@@ -4,25 +4,127 @@
...
 
-use dusk_jubjub::{JubJubAffine, JubJubScalar, GENERATOR};
+use dusk_jubjub::{JubJubAffine, JubJubScalar, GENERATOR, GENERATOR_EXTENDED};
+use ff::Field;
 use rand_core::OsRng;
 
-use phoenix_core::{decrypt, encrypt, ENCRYPTION_EXTRA_SIZE};
+use phoenix_core::{aes, elgamal, PublicKey, SecretKey};
+
+#[cfg(feature = "zk")]
```
moCello: you could add a 'tests/gadgets.rs' to avoid the `#[cfg(feature = "zk")` repetition.

moCello (commented):
I like how these `aes::encrypt` and `elgamal::decrypt` now come together :)

One comment on the naming side: I prefer to name `zk_encrypt` -> `encrypt_gadget` since we are using that convention across our stack.

And I was wondering why there is only a `zk_encrypt`? Are we not decrypting in the circuit?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ exclude = [".github/workflows/dusk-ci.yml", ".gitignore"]
 [dependencies]
 rand_core = { version = "0.6", default-features = false }
 dusk-bytes = "0.1"
+dusk-plonk = { version = "0.19", default-features = false, optional = true }
```
miloszm, line 14: If dusk-plonk is needed only in tests, should this dependency be included only as dev-dependency? I think it should but pls. correct me if there is a reason not to.

miloszm (dismissed):

Cargo.toml
xevisalle: plonk is not only used in tests, is used in the gadget module to encrypt in ZK

xevisalle:
> I like how these `aes::encrypt` and `elgamal::decrypt` now come together :)
> 
> One comment on the naming side: I prefer to name `zk_encrypt` -> `encrypt_gadget` since we are using that convention across our stack.
> 
> And I was wondering why there is only a `zk_encrypt`? Are we not decrypting in the circuit?

@marta-belles do you think it makes sense to have a decryption method in-circuit? I didn't implement it cause we don't need it, but I guess it doesn't hurt to have it just in case.

marta-belles:
> > I like how these `aes::encrypt` and `elgamal::decrypt` now come together :)
> > One comment on the naming side: I prefer to name `zk_encrypt` -> `encrypt_gadget` since we are using that convention across our stack.
> > And I was wondering why there is only a `zk_encrypt`? Are we not decrypting in the circuit?
> 
> @marta-belles do you think it makes sense to have a decryption method in-circuit? I didn't implement it cause we don't need it, but I guess it doesn't hurt to have it just in case.

I can't think of a case that we will prove decryption in-circuit, but there is no harm in having it there for completeness:)

src/encryption/elgamal.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Decrypts a ciphertext given a secret key,
+/// returning a JubJubExtended plaintext
+pub fn decrypt(
+    secret_key: &JubJubScalar,
+    ciphertext_1: &JubJubExtended,
+    ciphertext_2: &JubJubExtended,
+) -> JubJubExtended {
+    ciphertext_2 - ciphertext_1 * secret_key
```
marta-belles, line 38: If it's not too much trouble, for clarity I suggest to name the result as `plaintext`. That is, do something like:
```
let plaintext = ciphertext_2 - ciphertext_1 * secret_key
```

src/encryption/elgamal.rs
```diff
@@ -0,0 +1,60 @@
...
+    r: &JubJubScalar,
+) -> Result<(WitnessPoint, WitnessPoint), Error> {
+    // IMPORT INPUTS
+    let public_key = composer.append_point(*public_key);
+    let plaintext = composer.append_point(*plaintext);
+    let r = composer.append_witness(*r);
+
+    // ENCRYPT
+    let R = composer.component_mul_point(r, public_key);
+    let ciphertext_1 = composer.component_mul_generator(r, GENERATOR)?;
```
marta-belles, line 51: Sorry for my ignorance in rust, but why do we have `?` here and not in the other lines? Is it because this is the only operation that can return a potential error?

marta-belles (approved):

src/encryption/elgamal.rs
xevisalle: yes, this function we call is implemented so that can return an error, and the others don't

PR #164, opened 2024-04-30
Add contract deployment transaction

### LINKED ISSUE #163: Add contract deployment transaction
#### Summary
A transaction type should be added to deal with the deployment of contracts. This is necessary for a [contract deployment system](https://github.com/dusk-network/rusk/issues/1071) to be able to function in the context of phoenix.

The transaction type should allow for charging the transaction, as well as deploying a contract - meaning it should include all the relevant data and metadata to perform this. 

A possible way to tackle this is to replace the `Transaction::call` field by a new field - maybe called `exec` - that contains an enum that can either be a contract call, or a contract deployment.

```rust
struct Transaction {
    /* other fields */
    exec: Execution,
}

enum Execution {
    Call { /*fields*/ }
    Deploy { /*fields*/ }
}
```

#### Additional context
Upstream should be able to use the data of this transaction to charge the user, as well as make decisions on *how much* to charge. This means the fields of the new transaction should likely be public, or otherwise accessible.

ureeves:
The deployment transaction contains the metadata necessary to deploy a contract, together with its bytecode and possible constructor arguments.

Resolves: #163

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -12,6 +12,7 @@ pub use stake::*;
 mod transfer;
 pub use transfer::*;
 
+use alloc::boxed::Box;
 use alloc::string::String;
 use alloc::vec::Vec;
 
@@ -21,7 +22,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 use dusk_bls12_381::BlsScalar;
 use dusk_bytes::{DeserializableSlice, Error as BytesError, Serializable};
 
-use crate::{Crossover, Fee, Note};
+use crate::{Crossover, Fee, Note, PublicKey};
 
 /// Type alias for the ID of a module.
 pub type ModuleId = [u8; 32];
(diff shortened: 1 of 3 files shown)


moCello (commented):
Since we now have the `transaction-types` in the rusk contracts, wouldn't it make sense to add the `Transaction` there too? Ideally I would like to get rid of the `mod transaction` in phoenix-core all together.

Edit: I think that is part of a bigger discussion though and not strictly related to this PR.

src/transaction.rs
```diff
@@ -48,10 +49,40 @@ pub struct Transaction {
...
+        /// The function to call in the contract
+        fn_name: String,
+        /// The argument to the function to call
+        fn_arg: Vec<u8>,
+    },
+    /// A contract deployment.
+    Deploy {
+        /// The owner of the contract
+        owner: Box<PublicKey>, /* boxed due to a large size difference
+                                * between variants */
```
moCello, line 79: I'm not sure what you mean with the large size difference between variants. To me it seems that the `PublicKey`, with it's fixed 64 bytes is not that much bigger than the 32 bytes of the `ModuleId` and the rest of the fields are pointers. Am I missing something?

moCello (approved):
This LGTM
As far as I understand a `Transaction` can now be 3 different things:
- a normal transaction (`exec = None`)
- a call to a contract function (`exec = Some(call)`)
- a deployment of a new contract (`exec = Some(deploy)`)

With the changes that are coming up in phoenix, we might need to rethink much of the overall design of the contract but that's for later ;)

src/transaction.rs
ureeves: A `PublicKey` is 320 bytes in memory - there's two extended points in there. So between that and 32, there quite some bytes.

moCello: oh wait, I was thinking wrong, you are right!

src/transaction.rs
```diff
@@ -137,17 +185,42 @@ impl Transaction {
...
+                    bytes.extend(fn_name.as_bytes());
+
+                    bytes.extend(fn_arg);
+                }
+                Execution::Deploy {
+                    owner,
+                    bytecode,
+                    constructor_args,
+                } => {
+                    bytes.push(2);
```
miloszm, line 211: Since the encodings 0,1,2 are also used below for deserialisation, could we maybe introduce some implicit constants for them? Also, case 0 would be easier to understand if it had a name.

miloszm (changes requested):

ureeves:
Not relevant anymore

PR #192, opened 2024-05-31
circuits: Remove `ViewKey` from `TxOutputNote::new`

### LINKED ISSUE #191: circuits: Change the way we generate a `TxOutputNote`
#### Summary
To generate a `TxOutputNote` we currently need the `ViewKey` of the notes recipient. This `ViewKey` can not be generated for the change-note.

#### Possible Solution
Replace the `ViewKey` parameter with the `value` and `blinder` used to generate the output note.

moCello:
Resolves #191

### DIFF
--- a/circuits/src/transaction.rs
+++ b/circuits/src/transaction.rs
@@ -5,7 +5,8 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use dusk_jubjub::{
-    JubJubScalar, GENERATOR, GENERATOR_NUMS, GENERATOR_NUMS_EXTENDED,
+    JubJubAffine, JubJubScalar, GENERATOR, GENERATOR_EXTENDED, GENERATOR_NUMS,
+    GENERATOR_NUMS_EXTENDED,
 };
 use dusk_plonk::prelude::*;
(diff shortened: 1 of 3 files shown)


circuits/src/transaction.rs
```diff
@@ -137,16 +138,18 @@ struct WitnessTxOutputNote {
...
-            value: note.value(Some(vk))?,
-            value_commitment: note.value_commitment().into(),
-            blinding_factor: note.blinding_factor(Some(vk))?,
-        })
+    /// Crate a new `TxOutputNote`.
+    pub fn new(value: u64, blinding_factor: JubJubScalar) -> Self {
+        let value_commitment = JubJubScalar::from(value);
+        let value_commitment = (GENERATOR_EXTENDED * value_commitment)
+            + (GENERATOR_NUMS_EXTENDED * blinding_factor);
+        let value_commitment = JubJubAffine::from(value_commitment);
```
HDauven: Not sure if I'm a fan of all this variable shadowing here, but if we do this more often ignore this comment.

HDauven (approved):

circuits/src/transaction.rs
moCello: Good point, fixed it

PR #200, opened 2024-06-10
Add ownability trait

### LINKED ISSUE #146: `SecretKey` should be enough to verify note's ownership
#### Summary
We created the `ViewKey` as a way to delegate the ownership check, however we don't always do that: for example, in the current implementation of the wallet (CLI and web) we're verifying the note locally.
In order to do so, we need to create a `ViewKey` even if we have access to a `SecretKey`, that should be enough to verify the note's ownership.

As the `ViewKey` has, the `SecretKey` should implement the `owns` method as well.

#### Additional context
The hope is to reduce not only the code involved, but at least a bit of the operations too.

xevisalle:
Resolves: https://github.com/dusk-network/phoenix/issues/146

### DIFF
--- a/core/src/keys.rs
+++ b/core/src/keys.rs
@@ -26,3 +26,24 @@ pub trait Ownable {
     /// Returns the associated `StealthAddress`
     fn stealth_address(&self) -> stealth::StealthAddress;
 }
+
+pub(crate) fn owns_unchecked(
+    a: &JubJubScalar,
+    owner: &impl crate::Ownable,
+) -> bool {
+    let sa = owner.sync_address();
+    let aR = sa.R() * a;
+
+    sa.k() == &aR
+}
+
+/// This trait grants the ability to check if a note is owned.
+pub trait Ownability {
+    /// Checks if:
+    /// `note_pk ?= H(R · a) · G + B` for `ViewKey`
+    /// `note_pk ?= (H(R · a) + b) · G` for `SecretKey`
+    fn owns(&self, owner: &impl crate::Ownable) -> bool;
+
+    /// Checks if `k_sync ?= R_sync · a`
+    fn owns_unchecked(&self, owner: &impl crate::Ownable) -> bool;
+}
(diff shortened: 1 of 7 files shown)


core/src/keys.rs
```diff
@@ -26,3 +26,24 @@ pub trait Ownable {
...
+/// This trait grants the ability to check if a note is owned.
+pub trait Ownability {
+    /// Checks if:
+    /// `note_pk ?= H(R · a) · G + B` for `ViewKey`
+    /// `note_pk ?= (H(R · a) + b) · G` for `SecretKey`
+    fn owns(&self, owner: &impl crate::Ownable) -> bool;
+
+    /// Checks if `k_sync ?= R_sync · a`
+    fn owns_unchecked(&self, owner: &impl crate::Ownable) -> bool;
+}
```
moCello, line 49: I think you can get away with only requiring a `fn a` function that returns the internal `a`.
`owns` and `owns_unchecked` can by computed from that, which means that the type that implements the `Ownability` trait doesn't need to re-compute the checks themselves.

```suggestion
pub trait Ownability {
    fn a(&self) -> JubJubScalar;
    
    fn owns(&self, owner: &impl crate::Ownable) -> bool {
        // implement owns here directly
    }

    fn owns_unchecked(&self, owner: &impl crate::Ownable) -> bool {
        // implement owns_unchecked here directly
    }
}
```

moCello (changes requested):

core/src/keys.rs
moCello: With the `Ownability` trait we should also let the function `note.decrypt_data` (and with that also `note.value` and note.blinding_factor`) to take any type that implements `Ownability`.
This means that we can also decrypt the value and blinding factor with a `SecretKey`

xevisalle: @moCello regarding your first comment: but then, how am I supposed to get `b` or `B`? In the end I still need two different ways to perform the check.

Regarding the decrypt data thing, I can look into it in a new commit.

xevisalle:
Closing because `Ownable` is no longer there.

PR #209, opened 2024-06-13
Move elgamal encryption and `RecipientParameters` to `phoenix-core`

ureeves:
This is required to allow depending on `RecipientParameters` without depending on `phoenix-circuits` directly.


moCello (approved):
LGTM

PR #211, opened 2024-06-13
Add `RecipientParameters` to `TxSkeleton`

ureeves:
The `RecipientParameter` are needed in the `TxSkeleton` to create the public-inputs for the verifier.

### DIFF
--- a/core/tests/transaction.rs
+++ b/core/tests/transaction.rs
@@ -7,9 +7,11 @@
 #![cfg(feature = "alloc")]
 
 use dusk_bls12_381::BlsScalar;
-use dusk_jubjub::JubJubScalar;
+use dusk_jubjub::{JubJubScalar, GENERATOR};
 use ff::Field;
-use phoenix_core::{Error, Note, PublicKey, SecretKey, TxSkeleton};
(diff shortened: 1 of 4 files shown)


core/tests/transaction.rs
```diff
@@ -28,13 +30,15 @@ fn transaction_parse() -> Result<(), Error> {
     let outputs = [note.clone(), note];
     let tx_max_fee = 0;
     let deposit = 0;
+    let recipient_params = RecipientParameters::default();
```
moCello: I would like to test the serialization with values other than only zeros.
```suggestion
    let recipient_params = RecipientParameters::new(
        &mut rng,
        &sk,
        [
        note.stealth_address().note_pk().as_ref().into(),
        note.stealth_address().note_pk().as_ref().into()
        ],
        BlsScalar::from(42);
    );
```
you need to see whether the note_pk can be constructed like this, you might need to use `JubJubAffine::from` instead of `into`.

moCello (changes requested):
one comment

moCello (approved):
LGTM

PR #215, opened 2024-06-14
Add sender encryption to the `Note`

### LINKED ISSUE #214: Add sender encryption to the `Note`
#### Summary
The current implementation of the recipient parameter introduced a recursion downstream that required a complete refactor of the recipient gadget.
The problem was that the `RecipientParameters` included the payload-hash, also needed to be included in the transaction payload themselves.

#### Detailed Description
This includes:
- Remove `RecipientParameters`
- Add sender-encryption to the `Note`
- Restructure the recipient-gadget

moCello:
Resolves #214


circuits/CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add Recipient gadget [#197]
-- Add `RecipientParameter::new` constructor [#201]
 
 ### Changed
 
+- Rename `blincing_factor` to `value_blinder` [#214]
```
ureeves:
```suggestion
- Rename `blinding_factor` to `value_blinder` [#214]
```

circuits/src/transaction.rs
```diff
@@ -329,12 +330,16 @@ fn nullify_gadget<const H: usize, const I: usize>(
...
     payload_hash: BlsScalar,
     root: BlsScalar,
     deposit: u64,
     max_fee: u64,
-    rp: RecipientParameters,
+    sender_pk: PublicKey,
+    signatures: (SchnorrSignature, SchnorrSignature),
+    output_npk: [JubJubAffine; OUTPUT_NOTES],
+    sender_blinder: [(JubJubScalar, JubJubScalar); OUTPUT_NOTES],
+    // rp: RecipientParameters,
```
ureeves: Should this remain here? :thinking:

circuits/tests/transaction.rs
```diff
@@ -72,9 +78,22 @@ lazy_static! {
...
+        let sig_b = schnorr_sk_b.sign(&mut rng, payload_hash);
+
+    let sender_blinder_0 = (
+        JubJubScalar::random(&mut rng),
+        JubJubScalar::random(&mut rng),
+    );
+    let sender_blinder_1 = (
+        JubJubScalar::random(&mut rng),
+        JubJubScalar::random(&mut rng),
+    );
```
ureeves:
```suggestion
        let schnorr_sk_b = SchnorrSecretKey::from(sender_sk.b());
        let sig_b = schnorr_sk_b.sign(&mut rng, payload_hash);

        let sender_blinder_0 = (
            JubJubScalar::random(&mut rng),
            JubJubScalar::random(&mut rng),
        );
        let sender_blinder_1 = (
            JubJubScalar::random(&mut rng),
            JubJubScalar::random(&mut rng),
        );
```

circuits/tests/transaction.rs
```diff
@@ -129,6 +152,35 @@ fn create_test_tx_input_notes<const I: usize>(
     input_notes.try_into().unwrap()
 }
 
+fn create_tx_output_note(
+    rng: &mut (impl RngCore + CryptoRng),
+    value: u64,
+    note_pk: JubJubAffine,
+    // (blinder_a, blinder_b)
```
ureeves: Errant comment?

ureeves (approved):
Looks good, some nitty comments

circuits/tests/transaction.rs
moCello: Can be a helpful comment for the future

circuits/src/recipient.rs
```diff
@@ -6,60 +6,110 @@
...
+fn assert_sender_enc(
+    composer: &mut Composer,
+    sender_pk_A: WitnessPoint,
+    sender_pk_B: WitnessPoint,
+    note_pk: WitnessPoint,
+    blinder: (Witness, Witness),
+    sender_enc: [(JubJubAffine, JubJubAffine); 2],
+) -> Result<(), Error> {
+    let blinder_A = blinder.0;
+    let (enc_A_c0, enc_A_c1) =
```
xevisalle: I understand why you are changing this, I guess it's basically because in the note_pk you want to match the number with the array index. However, I have 2 cents here:
- I would try not to make changes when it's just a matter of preference
- In this particular example, where we are writing mathematical / cryptographic notation, I would stick with the common math / crypto notation. Although it's also a matter of preference, it's very common to start with 1. In this specific case, all resources you can find explaining ElGamal, even the original paper from 1984, use the `(c1, c2)` notation. As in the module is still c1, c2, I'd keep the notation

xevisalle (changes requested):
After discussing with @marta-belles , and after seeing the overall code, we think it makes more sense to call the `recipient` module / gadget / whatever, `sender`.

circuits/src/transaction.rs
```diff
@@ -131,27 +129,30 @@ impl<const H: usize> TxInputNote<H> {
...
+    value_blinder: JubJubScalar,
+    sender_enc: [(JubJubAffine, JubJubAffine); 2],
 }
 
 #[derive(Debug, Clone)]
 struct WitnessTxOutputNote {
     value: Witness,
     value_commitment: WitnessPoint,
-    blinding_factor: Witness,
+    value_blinder: Witness,
```
marta-belles: I see that `sender_enc` is part of `TxOutputNote` but not part of this struct, why is that?

circuits/src/transaction.rs
```diff
@@ -131,27 +129,30 @@ impl<const H: usize> TxInputNote<H> {
 pub struct TxOutputNote {
     value: u64,
     value_commitment: JubJubAffine,
-    blinding_factor: JubJubScalar,
+    value_blinder: JubJubScalar,
+    sender_enc: [(JubJubAffine, JubJubAffine); 2],
```
xevisalle, line 134: it feels odd to have `sender_enc` here and not in WitnessTxOutputNote. I don't have a strong preference, but I maybe would keep it as an independent parameter.

circuits/src/transaction.rs
```diff
@@ -131,27 +129,30 @@ impl<const H: usize> TxInputNote<H> {
...
 }
 
 impl TxOutputNote {
     /// Create a new `TxOutputNote`.
     pub fn new(
         value: u64,
         value_commitment: JubJubAffine,
-        blinding_factor: JubJubScalar,
+        value_blinder: JubJubScalar,
+        sender_enc: [(JubJubAffine, JubJubAffine); 2],
```
marta-belles, line 144: To me it feels that the `sender_enc` and the `npk` of the output notes should belong to the same struct. But if it's a big change then it doesn't really matter.

circuits/src/transaction.rs
```diff
@@ -329,12 +330,15 @@ fn nullify_gadget<const H: usize, const I: usize>(
...
+    tx_output_notes: [TxOutputNote; OUTPUT_NOTES],
     payload_hash: BlsScalar,
     root: BlsScalar,
     deposit: u64,
     max_fee: u64,
-    rp: RecipientParameters,
+    sender_pk: PublicKey,
+    signatures: (SchnorrSignature, SchnorrSignature),
+    output_npk: [JubJubAffine; OUTPUT_NOTES],
+    sender_blinder: [(JubJubScalar, JubJubScalar); OUTPUT_NOTES],
```
marta-belles, line 319: As before, it would make more sense to me to keep the `sender_blinder` together with the encryption, but if it's too much work, nevermind.

circuits/tests/elgamal.rs
```diff
@@ -113,7 +91,14 @@ fn test_elgamal_gadgets() {
     let (proof, public_inputs) = prover
         .prove(
             &mut rng,
-            &ElGamalCircuit::new(&pk.A(), &sk.a(), &message, &r, &c1, &c2),
+            &ElGamalCircuit::new(
+                &pk.A(),
+                &sk.a(),
+                &message,
+                &r,
+                &c1.into(),
```
marta-belles, line 99: Should these be 
```
&blinder,
&c0.into(),
&c1.into(),
```
to make it consistent  with the rest of the code?

circuits/tests/transaction.rs
marta-belles: Does this mean the blinders for the encryption of the sender's `A` and `B`? If so, maybe change it to `(blinder_A, blinder_B)`?

circuits/src/recipient.rs
moCello: I was mostly changing the indexes for the output notes to match the indices in the output-notes array. I can change the once for the the elgamal ciphertext back if you want.

circuits/src/transaction.rs
moCello: They are appended in the `sender_gadget` so appending them here is not needed and would in fact be double.

core/src/transaction.rs
```diff
@@ -37,8 +37,6 @@ pub struct TxSkeleton {
     pub tx_max_fee: u64,
```
marta-belles: Wasn't this variable renamed to `max_fee`? Another thing, `outputs` includes the encryptions, right?

core/tests/encryption.rs
```diff
@@ -28,3 +29,25 @@ fn test_aes_encrypt_and_decrypt() {
...
+fn test_elgamal_encrypt_and_decrypt() {
+    let mut rng = StdRng::seed_from_u64(0xc0b);
+
+    let sk = SecretKey::random(&mut rng);
+    let pk = PublicKey::from(&sk);
+
+    let message = GENERATOR_EXTENDED * JubJubScalar::from(1234u64);
+
+    // Encrypt using a fresh random value 'r'
+    let r = JubJubScalar::random(&mut rng);
```
marta-belles: It's a test and a different scope, so it doesn't really matter but, would it be better to change `r` to `blinder` and `(c1, c2)` to `(c0, c1)` to make it consistent with the rest?

xevisalle (approved):
LGTM

circuits/src/transaction.rs
moCello: We are appending the `sender_enc` in the `sender_enc::gadget`, appending it here would add 4 more fields to the `WitnessTxOutputNote` struct.

circuits/tests/elgamal.rs
moCello: I ended up using the `c1` and `c2` notation, so no need to change anything here

circuits/src/transaction.rs
moCello: That makes a lot of sense actually. I'll change

circuits/src/transaction.rs
moCello: I decided to remove the `append_to_circuit` method all together and let the different gadgets append the fields as they need them.

PR #216, opened 2024-06-17
Add `owns` and `owns_unchecked` to `SecretKey`

### LINKED ISSUE #146: `SecretKey` should be enough to verify note's ownership

xevisalle:
Resolves: https://github.com/dusk-network/phoenix/issues/146


HDauven (approved):
LGTM, +1 for the clear variable naming

PR #218, opened 2024-06-18
core: Add sender public key to note costructors

### LINKED ISSUE #214: Add sender encryption to the `Note`

moCello:
Resolves #214

### DIFF
--- a/core/src/encryption/elgamal.rs
+++ b/core/src/encryption/elgamal.rs
@@ -33,9 +33,11 @@ pub fn encrypt(
 /// Returns a JubJubExtended plaintext.
 pub fn decrypt(
     secret_key: &JubJubScalar,
-    ciphertext_1: &JubJubExtended,
-    ciphertext_2: &JubJubExtended,
+    ciphertext: &(JubJubExtended, JubJubExtended),
 ) -> JubJubExtended {
+    let ciphertext_1 = ciphertext.0;
+    let ciphertext_2 = ciphertext.1;
+
     // return the plaintext
     ciphertext_2 - ciphertext_1 * secret_key
 }
--- a/core/src/note.rs
+++ b/core/src/note.rs
@@ -6,19 +6,18 @@
 
 use core::convert::{TryFrom, TryInto};
 
-use crate::{
-    encryption::elgamal, transparent_value_commitment, value_commitment, Error,
-    PublicKey, SecretKey, StealthAddress, ViewKey,
-};
 use dusk_bls12_381::BlsScalar;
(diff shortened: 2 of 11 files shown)


core/src/encryption/elgamal.rs
```diff
@@ -33,9 +33,11 @@ pub fn encrypt(
 /// Returns a JubJubExtended plaintext.
 pub fn decrypt(
     secret_key: &JubJubScalar,
-    ciphertext_1: &JubJubExtended,
-    ciphertext_2: &JubJubExtended,
+    ciphertext: &(JubJubExtended, JubJubExtended),
```
xevisalle, line 36: this is a breaking change, should be added to the changelog.

Plus, I think the change itself is just a matter of preference, I'd try to avoid this kind of changes.

core/src/note.rs
```diff
@@ -96,13 +94,14 @@ impl Note {
     pub fn new<R: RngCore + CryptoRng>(
         rng: &mut R,
         note_type: NoteType,
-        pk: &PublicKey,
+        sender_pk: &PublicKey,
+        receiver_pk: &PublicKey,
```
xevisalle, line 98: they told me "receiver" is not correct in English? Isn't it "recipient"?

HDauven (approved):

xevisalle (approved):
LGTM, just add that change to the changelog, and double-check the "receiver" wording

core/src/note.rs
HDauven: > they told me "receiver" is not correct in English? Isn't it "recipient"?

Why would it not be correct? I do agree `recipient` is better though, it's more formal

core/src/encryption/elgamal.rs
moCello: I changed it so the decrypt is the direct inverse of the encrypt, so to me that counts for more than just preference ;)

moCello: The changelog entry is already under the unrelease section:
```
- Add `elgamal::encrypt` and `elgamal::decrypt`
```

moCello: There was no release in the meantime so the change is also not breaking (within core)

PR #220, opened 2024-06-19
core: Remove `SyncAddress`

### LINKED ISSUE #219: Remove `SyncAddress`
#### Summary

Due to recent findings about the performance of wallet synchronization in https://github.com/dusk-network/phoenix-benchmark/pull/1 we figure it would be wiser to remove the `SyncAddress` from a `Note`. It has some unsavory API costs that we would prefer to not incur. 

This should be a simple "remove and follow the consequences" job.

moCello:
Resolves #219


xevisalle (approved):
AIRIILGTM

HDauven (approved):

PR #223, opened 2024-06-25
core: Add `Sender` struct

### LINKED ISSUE #222: Add the option to set a contract as a sender of a `Note`
#### Summary
Not all `Note`s are created as a phoenix-transaction's output-note. Some will be created by e.g. withdrawing from a contract, withdrawing stake-rewards or as a genesis-`Note`. In these instances the *sender* is not a `PublicKey` but a contract.
To accommodate for such `Note`s the sender-field should be either the encryption of the `PublicKey` encryption (in case of a phoenix-output-note) or an array of bytes that can identify the contract with some additional metadata (such as the `BlsPublicKey` is the case of the stake-contract) so that the `Note`'s origin can be tracked.

moCello:
Resolves #222


core/src/note.rs
```diff
@@ -454,30 +410,164 @@ impl Serializable<SIZE> for Note {
...
+    /// Information to identify the origin of a `Note`, if it wasn't created as
+    /// a phoenix-transaction output-note.
+    ContractInfo([u8; 4 * JubJubAffine::SIZE]),
+}
+
+impl Sender {
+    /// Create a new [`Sender`] enum by encrypting the sender's [`PublicKey`] in
+    /// a way that only the receiver of the note can decrypt.
+    pub fn encrypt(
+        note_pk: &NotePublicKey,
```
xevisalle, line 461: I'd call this `recipient_note_pk`

xevisalle (approved):

core/src/note.rs
moCello: Do you insist? Because I think it adds more noise. In the end the npk of the stealth-address is used.

xevisalle: I do not!

core/src/note.rs
```diff
@@ -454,30 +410,164 @@ impl Serializable<SIZE> for Note {
...
+    derive(Archive, Serialize, Deserialize),
+    archive_attr(derive(bytecheck::CheckBytes))
+)]
+pub enum Sender {
+    /// The sender's [`PublicKey`], encrypted using the note_pk of the
+    /// stealth-address.
+    Encryption([(JubJubAffine, JubJubAffine); 2]),
+    /// Information to identify the origin of a `Note`, if it wasn't created as
+    /// a phoenix-transaction output-note.
+    ContractInfo([u8; 4 * JubJubAffine::SIZE]),
```
ureeves, line 454: I'm not sure I understand the usage of `JubJubAffine::SIZE` here. What is the intent?

ureeves (changes requested):

core/src/note.rs
moCello: The contract info is supposed to be the same size as four `JubJubAffine` would take otherwise.

moCello: This way the sender is always constant size and we can implement `Serializable` on it.

ureeves: I see, so its supposed to take the same size as serialized on the wire, and not in memory?

ureeves: > This way the sender is always constant size and we can implement `Serializable` on it.

Ah got it!

ureeves (approved):

PR #224, opened 2024-06-25
circuit: Make `TxInputNote` fields public

moCello:
Expose the fields of `TxInputNote` so it can also be constructed from the internal fields, without the sender-sk.

### DIFF
--- a/circuits/src/transaction.rs
+++ b/circuits/src/transaction.rs
@@ -160,158 +160,6 @@ impl TxOutputNote {
     }
 }
 
-/// Transaction gadget proving the following properties in ZK for a generic
-/// `I` [`TxInputNote`] and [`OUTPUT_NOTES`] (2) [`TxOutputNote`]:
-///
-/// 1. Membership: every [`TxInputNote`] is included in the Merkle tree of
-///    notes.
-/// 2. Ownership: the sender holds the note secret key for every
-///    [`TxInputNote`].
-/// 3. Nullification: the nullifier is calculated correctly.
-/// 4. Minting: the value commitment for every [`TxOutputNote`] is computed
-///    correctly.
-/// 5. Balance integrity: the sum of the values of all [`TxInputNote`] is equal
-///    to the sum of the values of all [`TxOutputNote`] + the gas fee + a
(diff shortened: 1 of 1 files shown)


circuits/src/transaction.rs
```diff
@@ -408,9 +263,44 @@ impl<const H: usize, const I: usize> TxCircuit<H, I> {
             sender_blinder,
         }
     }
+
+    /// Return the public-input values used for the circuit
+    pub fn public_inputs(&self) -> Vec<BlsScalar> {
```
xevisalle: why is this needed if that's returned by the prover?

circuits/src/transaction.rs
```diff
@@ -425,18 +315,132 @@ impl<const H: usize, const I: usize> Circuit for TxCircuit<H, I> {
...
-        nullify_gadget::<H, I>(
-            composer,
-            &payload_hash,
-            &self.root,
-            &self.tx_input_notes,
-            &self.tx_output_notes,
-            self.max_fee,
-            self.deposit,
-        )?;
+        // Append the root as public input
```
xevisalle, line 299: which is the advantage of adding this here and changing the already peer-reviewed comments?

circuits/src/transaction.rs
moCello: because in practice the verifier doesn't have access to the data that the prover returns.
To be fair, I'm also not sure if the verifier will use this function downstream or if the pi will be generated from the transaction itself. I actually think the latter is the case.

circuits/src/transaction.rs
moCello: Ease of work: the other file was a real pain to debug and I changed only the comments that directly refer to the steps 1 - 6 in the documentation of the function.
I also thought that it is good to have the circuit implementation be more close to the circuit outlined in the specs with clear markers where what things are checked (e.g. membership, ownership, etc)

circuits/src/transaction.rs
moCello: This got me thinking, I'll remove the function again!

ureeves (approved):
LGTM!

PR #225, opened 2024-06-26
core: Change `owns` to take a `StealthAddress`

moCello:
These changes are required after the deletion of the `Ownable` trait, in order for the functions to take both `Note`s and `License`s

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Let `owns` take a `StealthAddress` instead of a `Note`
 - Rename `tx_max_fee` to `max_fee` [#214]
 - Add `sender_enc` field to the `Note` [#214]
 - Add `sender_blinder` parameter for `Note` contructors [#214]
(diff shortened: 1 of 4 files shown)


xevisalle (commented):
it looks good to me, but I have two questions:
- since the code with `note` still works, why the change is needed?
- if the change is needed, maybe `sk.owns(sa)` feels weird? `sk.check_sa(sa)` or similar is more appropiate?

HDauven (approved):

HDauven:
> * if the change is needed, maybe `sk.owns(sa)` feels weird? `sk.check_sa(sa)` or similar is more appropiate?

I prefer `sk.owns(sa)` because it clearly conveys the relationship between the secret key and the ownership of the stealth address. It is much more declarative.

HDauven:
> * if the change is needed, maybe `sk.owns(sa)` feels weird? `sk.check_sa(sa)` or similar is more appropiate?

I prefer sk.owns(sa) because it clearly conveys the relationship between the secret key and the ownership of the stealth address. It is much more declarative.

moCello:
> it looks good to me, but I have two questions:
> 
>     * since the code with `note` still works, why the change is needed?
Because it doesn't work with `License`s.

>     * if the change is needed, maybe `sk.owns(sa)` feels weird? `sk.check_sa(sa)` or similar is more appropiate?
This would create more unneeded changes downstream that I like to avoid at this point.

ureeves (approved):

xevisalle (approved):

PR #233, opened 2024-08-08
circuits: Add `TxCircuit::from_slice` constructor

### LINKED ISSUE #232: circuits: Add a way to construct a `TxCircuit` from bytes
#### Summary
Since the `TxCircuit` will replace the `UnprovenTransaction` struct in the `rusk-prover` we need a way to create a `TxCircuit` from a blob of bytes.

We can achieve this in two ways:
- Implement the `Serializabla` trait on `TxCircuit` to be able to create the struct from bytes.
- Implement a `from_slice` constructor

The implementation of `Serializable` would be cleaner, but since it would also need to be generic over the height of the notes tree and the amount of input-notes it would probably require an implementation with a macro which seems overkill for this purpose.

#### Relevant Context
See [rusk #2069](https://github.com/dusk-network/rusk/issues/2069) and #229

moCello:
Resolves #232

### DIFF
--- a/circuits/CHANGELOG.md
+++ b/circuits/CHANGELOG.md
@@ -22,6 +22,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rename `TxInputNote` to `InputNoteInfo` [#229]
 - Rename `TxOutputNote` to `OutputNoteInfo` [#229]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


xevisalle (approved):
LEGTM

ureeves (approved):
LGTM!

PR #236, opened 2024-08-09
circuit: Add plonk dependency under default feature

### LINKED ISSUE #235: circuits: Add default feature to expose the circuit structs under no default-features
#### Summary
Add `default` feature so that we can use `TxCircuit`, `InputNoteInfo` and `OutputNoteInfo` with no default-feature downstream without immediately adding the entire dependency tree (plonk).

moCello:
Resolves #235


ureeves (approved):
Nice! LGTM

PR #241, opened 2024-08-13
core: Fix panic when decrypting note with incorrect view-key

### LINKED ISSUE #240: core: Fix panic when trying to decrypt the value for a note with an incorrect viewkey
**Describe the bug**
When attempting to decrypt the value for an obfuscated note with a view-key that doesn't own the note, the code doesn't error but panics due to entering unreachable code.

**To Reproduce**
In tests/note_test add:
```rust
#[test]
fn note_not_owned() {
    let mut rng = StdRng::seed_from_u64(0xc0b);

    let owner_pk = PublicKey::from(&SecretKey::random(&mut rng));
    let value_blinder = JubJubScalar::random(&mut rng);
    let sender_blinder = [
        JubJubScalar::random(&mut rng),
        JubJubScalar::random(&mut rng),
    ];

    let note = Note::obfuscated(
        &mut rng,
        &owner_pk,
        &owner_pk,
        42,
        value_blinder,
        sender_blinder,
    );

    let not_owner_sk = SecretKey::random(&mut rng);
    let not_owner_vk = ViewKey::from(&not_owner_sk);

    assert!(note.value(Some(&not_owner_vk)).is_err());
}
```

and run the test:
```
cargo test --tests note_not_owned
```

**Expected behaviour**
In case of an error we should gracefully return the error and not panic.

moCello:
Resolves #240

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update `bls12_381-bls` dep to 0.4
 
+### Fixed
+
+- Fix panic when attempting to decrypt the note with an incorrect view-key [#240]
+
 ## [0.30.0] - 2024-07-03
 
 ### Added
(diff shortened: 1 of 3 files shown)


HDauven (approved):
LGTM

PR #257, opened 2024-12-09
Added serde support

d-sonuga:
For dusk-network/rusk/issues/2773 and #258 .

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -22,7 +22,7 @@ jobs:
     name: test core
     uses: dusk-network/.github/.github/workflows/run-tests.yml@main
     with:
-      test_flags: -p phoenix-core --features=alloc
+      test_flags: -p phoenix-core --features=alloc,serde
 
   test_core_rkyv:
     name: test core rkyv compiles
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -22,6 +22,11 @@ aes-gcm = { version = "0.10", default-features = false, features = ["aes", "allo
 zeroize = { version = "1", default-features = false, features = ["derive"] }
 rkyv = { version = "0.7", optional = true, default-features = false }
 bytecheck = { version = "0.6", optional = true, default-features = false }
+serde = { version = "1.0", optional = true }
+bs58 = { version = "0.4" , optional = true }
+base64 = { version = "0.22", optional = true }
+serde_json = { version = "1.0", optional = true }
+hex = { version = "0.4" , optional = true }
 
 [dev-dependencies]
 assert_matches = "1.3"
@@ -39,3 +44,4 @@ rkyv-impl = [
     "rkyv",
     "bytecheck"
 ]
+serde = ["dep:serde", "bs58", "base64", "hex", "serde_json", "dusk-jubjub/serde"]
--- a/core/src/serde_support.rs
+++ b/core/src/serde_support.rs
@@ -1,0 +1,589 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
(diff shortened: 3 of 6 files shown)


core/src/serde_support.rs
```diff
@@ -0,0 +1,589 @@
...
+            SerdeError::invalid_length(decoded_len, &byte_length_str.as_str())
+        })?;
+        StealthAddress::from_bytes(&bytes)
+            .map_err(|err| SerdeError::custom(format!("{err:?}")))
+    }
+}
+
+// To serialize and deserialize u64s as big ints:
+// https://github.com/dusk-network/rusk/issues/2773#issuecomment-2519791322.
+struct Bigint(u64);
```
Neotamandua, line 169: Is this also needed in other repositories?

core/src/serde_support.rs
```diff
@@ -0,0 +1,589 @@
...
+                })
+            }
+        }
+
+        deserializer.deserialize_struct("Note", &FIELDS, NoteVisitor)
+    }
+}
+
+// The current serde implementation for `BlsScalar` is not what it's expected to
+// be, so this is needed at the moment.
```
Neotamandua: Could you add the issue: https://github.com/dusk-network/bls12_381/issues/145 to this comment as well?

core/src/serde_support.rs
d-sonuga: The other repos' structs didn't have any `u64`s to serialize.

core/src/serde_support.rs
```diff
@@ -0,0 +1,589 @@
...
+
+impl<'de> Deserialize<'de> for Bigint {
+    fn deserialize<D: Deserializer<'de>>(
+        deserializer: D,
+    ) -> Result<Self, D::Error> {
+        let mut s = String::deserialize(deserializer)?;
+        let last_char = s.pop().ok_or_else(|| {
+            SerdeError::invalid_value(
+                Unexpected::Str(&s),
+                &"a bigint ending with character 'n'",
```
moCello: I think here the error should say that we are expecting the string not to be empty.

Neotamandua (changes requested):
We also want to run this again in the CI, so that the tests will be executed as well in the future.

.github/workflows/dusk_ci.yml
```diff
@@ -22,13 +22,13 @@ jobs:
...
     with:
-      test_flags: -p phoenix-core --features=alloc
+      test_flags: -p phoenix-core --features=alloc,serde
 
   test_core_rkyv:
     name: test core rkyv compiles
     uses: dusk-network/.github/.github/workflows/run-tests.yml@main
     with:
-      test_flags: -p phoenix-core --features=rkyv-impl,alloc --no-run
+      test_flags: -p phoenix-core --features=rkyv-impl,alloc,serde --no-run
```
moCello: This test-run just wants to make sure that everything compiles when the `rkyv-impl` feature is enabled.
Since we already know that the code compiles with the `serde` feature enabled (due to the above `test core` test-run) and the `serde` feature has no connection to the `rkyv-impl` feature, I don't think that we need to add the `serde` feature flag to this test-run.

core/Cargo.toml
```diff
@@ -39,3 +44,4 @@ rkyv-impl = [
     "rkyv",
     "bytecheck"
 ]
+serde = ["dep:serde", "dusk-bls12_381/serde", "bs58", "base64", "hex", "serde_json", "dusk-jubjub/serde"]
```
moCello: please don't use `dusk-bls12_381/serde` because it doesn't use the serde serialization we want (and changing it would result in a major version bump).
@Neotamandua has a workaround for serializing the `BlsScalar`

moCello (approved):
Looks good from my side but please make sure that @Neotamandua is also on board with all the changes

Neotamandua (approved):
Looks good to me too. The only thing I would have suggested is to add a similar test as in piecrust to catch the case where a struct field is changed, but I don't think that's really needed here, so let's leave it as is.

PR #260, opened 2024-12-17
Compute the AES key using HKDF

### LINKED ISSUE #261: core: Improve key derivation for AES
#### Summary
We should convert the shared secret key (a JubJubAffine) computed through ECDH to a suitable symmetric key for AES encryption. We should use HKDF for the key derivation. To be precise, we should use SHA256, with (for instance) the Note's stealth address as a Salt, and the label "Phoenix-Dusk" to bound this to our protocol.

Reference: https://datatracker.ietf.org/doc/html/rfc5869

xevisalle:
Resolves: #261

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Compute AES key using HKDF [#261]
+
 ## [0.32.1] - 2024-12-17
 
 ### Added
@@ -21,7 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Call `gen_note_sk` in `SecretKey::owns` to avoid code duplication [#246]
 - `ViewKey` now checks both `a` and `B` in `ct_eq()`
-- Compute AES key using HKDF
 
 ### Removed
 
@@ -403,6 +406,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Canonical implementation shielded by feature.
 
 <!-- ISSUES -->
+[#261]: https://github.com/dusk-network/phoenix/issues/261
 [#258]: https://github.com/dusk-network/phoenix/issues/258
 [#255]: https://github.com/dusk-network/phoenix/issues/255
 [#240]: https://github.com/dusk-network/phoenix/issues/240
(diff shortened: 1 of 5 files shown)


marta-belles (commented):
It looks good to me, but I think it's best if a dev also takes a look at it.

core/Cargo.toml
```diff
@@ -27,6 +27,9 @@ bs58 = { version = "0.4" , optional = true }
 base64 = { version = "0.22", optional = true }
 serde_json = { version = "1.0", optional = true }
 hex = { version = "0.4" , optional = true }
+hkdf = "0.12.4"
+sha2 = "0.10.8"
```
Neotamandua:
```suggestion
hkdf = "0.12"
sha2 = "0.10"
```

Instead of specifying a particular subversion, I think we can change that to the major version only, as with other dependencies in the Cargo.toml

core/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Call `gen_note_sk` in `SecretKey::owns` to avoid code duplication [#246]
 - `ViewKey` now checks both `a` and `B` in `ct_eq()`
+- Compute AES key using HKDF
```
moCello: the issue link is missing

moCello (approved):
LGTM

core/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Call `gen_note_sk` in `SecretKey::owns` to avoid code duplication [#246]
 - `ViewKey` now checks both `a` and `B` in `ct_eq()`
+- Compute AES key using HKDF [#261]
```
Neotamandua: This should be in the unreleased section, but we can also change the CHANGELOG in today's release PR accordingly.

Neotamandua (approved):
LGTM

core/Cargo.toml
```diff
@@ -27,6 +27,9 @@ bs58 = { version = "0.4" , optional = true }
 base64 = { version = "0.22", optional = true }
 serde_json = { version = "1.0", optional = true }
 hex = { version = "0.4" , optional = true }
+hkdf = "0.12"
+sha2 = "0.10"
+hex-literal = "0.4"
```
herr-seppia: Is this required or just for tests?

core/CHANGELOG.md
xevisalle: my bad, fixing it now!

core/Cargo.toml
xevisalle: good point, it was used while developing for testing, it's not used atm. Removed.

PR #269, opened 2025-02-07
Update dependencies to versions with serde features & Remove `BlsScalar` serde newtype implementation

d-sonuga:


Neotamandua:
CI doesn't catch clippy errors on the serde feature. I suggest we add "--features=alloc,serde" to the clippy_args of the code_analysis job

Neotamandua (approved):

PR #286, opened 2026-06-15
Release Phoenix core 0.35.1 and circuits 0.8.1

HDauven:
## Summary

- merge Phoenix private maintenance history through `5d1934dcff8a4ba9bd365b8a6772a596681b1500`
- includes the Rusk 1.7 pinned private commit `68d48a1d6ad124bfd17b6d4770f6308778422d02`
- release `phoenix-core 0.35.1` and `phoenix-circuits 0.8.1`
- update circuits to depend on `phoenix-core 0.35.1` and `dusk-plonk 0.22.1`

## Notes

This keeps Rusk on the Phoenix 0.35/0.8 line while replacing the private git dependency with public patch releases.

## Validation

- `cargo package -p phoenix-core --allow-dirty --no-verify`
- `make cq`
- `make test`
- `make test-no-std`
- `make no-std`

`phoenix-circuits` packaging must be rerun after `phoenix-core 0.35.1` is published, because crates.io cannot resolve that dependency before publication.

### DIFF
--- a/core/src/transaction.rs
+++ b/core/src/transaction.rs
@@ -10,11 +10,10 @@
 extern crate alloc;
 use alloc::vec::Vec;
 
-#[cfg(feature = "rkyv-impl")]
-use rkyv::{Archive, Deserialize, Serialize};
-
 use dusk_bls12_381::BlsScalar;
 use dusk_bytes::{DeserializableSlice, Error as BytesError, Serializable};
+#[cfg(feature = "rkyv-impl")]
+use rkyv::{Archive, Deserialize, Serialize};
 
 use crate::{Note, OUTPUT_NOTES};
 
--- a/core/src/transaction.rs
+++ b/core/src/transaction.rs
@@ -97,8 +97,24 @@ impl TxSkeleton {
         let mut buffer = buf;
         let root = BlsScalar::from_reader(&mut buffer)?;
 
-        let num_nullifiers = u64::from_reader(&mut buffer)?;
-        let mut nullifiers = Vec::with_capacity(num_nullifiers as usize);
+        let num_nullifiers_u64 = u64::from_reader(&mut buffer)?;
+        let min_tail_len = OUTPUT_NOTES
+            .checked_mul(Note::SIZE)
+            .and_then(|len| len.checked_add(u64::SIZE + u64::SIZE))
+            .ok_or(BytesError::InvalidData)?;
+        if buffer.len() < min_tail_len {
+            return Err(BytesError::InvalidData);
+        }
+
+        let max_nullifiers_by_len =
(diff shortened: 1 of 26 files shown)


xevisalle (approved):

core/src/transaction.rs
```diff
@@ -94,32 +141,13 @@ impl TxSkeleton {
...
-            root,
-            nullifiers,
-            outputs,
-            max_fee,
-            deposit,
-        })
+    /// Deserialize the transaction from a bytes buffer, accepting historical
+    /// on-curve `JubJub` points inside output notes.
+    pub fn from_slice_legacy_compat(buf: &[u8]) -> Result<Self, BytesError> {
+        Self::from_slice_with(buf, Note::read_legacy_compat)
```
moCello, line 150: **NOTE:** New public legacy-compat decoders ship with no in-crate test, and are byte-identical to the strict path at the pinned `dusk-jubjub = "0.15"`, so their compat behavior is neither exercised nor pinned.

**Context:** Strict (`JubJubAffine::from_slice` → `Serializable::from_bytes`) and legacy (`affine_from_slice_legacy_compat` → inherent `JubJubAffine::from_bytes`) both perform on-curve + canonical checks only — neither calls `is_prime_order`/`is_torsion_free`, and the two are equivalent at `dusk-jubjub 0.15`. The legacy path is forward-compat scaffolding that diverges from strict only once `dusk-jubjub` adds a subgroup check to its decode; the `"0.15"` requirement won't admit that version. Same applies to `Note::`/`Sender::`/`StealthAddress::from_bytes_legacy_compat`. The only added test covers the nullifier bound on strict `from_slice`.
**Follow-up (non-blocking — tracked as a separate issue, not a merge gate):** Add an in-crate test decoding a `TxSkeleton`/`Note` byte vector through both `from_slice` and `from_slice_legacy_compat` (and the `Sender`/`StealthAddress` legacy decoders directly), asserting the documented relationship — equal today, divergent once the dependency carries the subgroup check. The test also documents that the legacy path is currently dormant.

moCello (approved):
APPROVE — Port of already-merged, reviewed private commits to the public repo. Version bumps, changelog formatting, the CI/Makefile migration, and the nullifier-allocation bound are correct and verified. The new legacy decoders read fields in the same order/sizes as the strict path (no corruption risk), are byte-identical to strict at the pinned `dusk-jubjub = "0.15"` (no subgroup check in either) — dormant forward-compat scaffolding, honestly described in the changelog — and are genuinely consumed downstream by rusk-core. No blockers, no correctness defects. One non-blocking observation: these decoders ship without an in-crate test pinning the strict-vs-legacy relationship; tracked as a follow-up issue rather than gating this port.

PR #289, opened 2026-07-11
Harden tx decode

### LINKED ISSUE #288: Harden strict tx decode: exact-length TxSkeleton + stealth membership
Phoenix transaction decoding accepts some non-canonical encodings: a length-prefixed `TxSkeleton` is not required to consume its full declared length, and `StealthAddress` deserialization admits identity / non-prime-order points.

These checks change what counts as a valid transaction, so they are consensus-observable and must activate atomically at a coordinated fork. The existing decoders are **not** tightened in place — they replay already-finalized blocks, and tightening them would reject that history. Instead the checks live on new `*_checked` decode entrypoints; the node routes the next-fork transaction format to them at activation.

Acceptance: new checked decoders that (a) require the `TxSkeleton` sub-region to be fully consumed and (b) reject identity / non-prime-order stealth points; existing decoders left byte-identical; regression tests for the rejected cases. Node-side routing of the new format to the checked decoders is tracked separately.

moCello:
Resolves #288

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `StealthAddress::from_bytes_checked`, rejecting non-prime-order stealth
+  points [#288]
+
 ### Changed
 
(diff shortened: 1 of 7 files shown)


HDauven (approved):
LGTM

xevisalle (approved):
lgtm

PR #291, opened 2026-08-05
core: Validate note openings against their commitment

### LINKED ISSUE #290: core: Validate decrypted note values against their commitment
#### Summary

Phoenix notes store both a value commitment and an encrypted value and blinder. We currently decrypt and trust the encrypted opening without checking that it matches the commitment.

A sender can therefore create a valid transaction where the proof uses one value and blinder for the commitment, while the encrypted opening contains different values. This does not inflate the supply, but wallets can display an incorrect balance and select a note they cannot spend.

Add a single method that decrypts the value and blinder, recomputes the commitment and returns an error if it does not match the commitment stored in the note.

`Note::value` and `Note::value_blinder` should use this checked path so existing consumers cannot accidentally trust an invalid opening.

Add tests for valid notes and mismatches in the value, blinder and both.

This should not require changes to the note format or circuit. Proving correct encryption in the circuit can be considered separately.

HDauven:
resolves #290

### DIFF
--- a/core/src/note.rs
+++ b/core/src/note.rs
@@ -102,6 +102,9 @@ impl PartialEq for Note {
 
 impl Note {
     /// Creates a new phoenix output note
+    ///
+    /// For transparent notes, `value_blinder` is ignored and the canonical
+    /// zero blinder is used instead.
     pub fn new<R: RngCore + CryptoRng>(
         rng: &mut R,
         note_type: NoteType,
@@ -114,6 +117,10 @@ impl Note {
         let r = JubJubScalar::random(&mut *rng);
         let stealth_address = receiver_pk.gen_stealth_address(&r);
 
+        let value_blinder = match note_type {
+            NoteType::Transparent => TRANSPARENT_BLINDER,
+            NoteType::Obfuscated => value_blinder,
+        };
         let value_commitment = value_commitment(value, value_blinder);
 
         // Output notes have undefined position, equals to u64's MAX value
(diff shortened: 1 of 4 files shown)


xevisalle (approved):
Not sure about the methods naming, but I don't have a strong opinion, so, lgtm. Nice improvement btw.

HDauven:
> Not sure about the methods naming, but I don't have a strong opinion, so, lgtm. Nice improvement btw.

I'm opening to changing it. #292 is a follow-up to prevent it from the other side, but I'm still trying to figure out the best/least painful approach.

core/src/note.rs
```diff
@@ -114,6 +114,10 @@ impl Note {
         let r = JubJubScalar::random(&mut *rng);
         let stealth_address = receiver_pk.gen_stealth_address(&r);
 
+        let value_blinder = match note_type {
+            NoteType::Transparent => TRANSPARENT_BLINDER,
+            NoteType::Obfuscated => value_blinder,
+        };
```
moCello, line 123: **WARNING:** `Note::new` now silently discards the caller's `value_blinder` for transparent notes — an observable behavior change that neither the doc comment nor the changelog mentions.

**Context:** The normalization is needed — without it, a transparent note built with a nonzero blinder would fail the new commitment check on every read — so the fix is to document it, not revert it. But the same call previously produced a commitment computed with the caller's blinder; now the argument is ignored, with no signal from docs, changelog, or the API itself.

**Fix:** State in the `Note::new` doc comment that `value_blinder` is ignored for transparent notes (the zero blinder is always used), and add a changelog line under Changed for this normalization.

core/src/note.rs
```diff
@@ -407,37 +411,44 @@ impl Note {
...
+    /// recovered value and blinder must reproduce the note's public value
+    /// commitment.
+    pub fn value_opening(
+        &self,
+        vk: Option<&ViewKey>,
+    ) -> Result<(u64, JubJubScalar), Error> {
+        let opening = match (self.note_type, vk) {
             (NoteType::Transparent, _) => {
                 let value =
                     u64::from_slice(&self.value_enc[..u64::SIZE]).unwrap();
```
moCello: **NOTE:** `unwrap()` in library code, carried over from the old `value` implementation into the rewritten `value_opening`.

**Context:** The slice is exactly `u64::SIZE` bytes so the decode cannot fail today, but the function already returns `Result<_, Error>` and `From<dusk_bytes::Error> for Error` exists — `?` works directly.

**Fix:** Replace `.unwrap()` with `?`.

core/src/note.rs
```diff
@@ -407,37 +411,44 @@ impl Note {
...
-                self.decrypt_value(vk).map(|(value, _)| value)
-            }
-            _ => Err(Error::MissingViewKey),
+            (NoteType::Obfuscated, Some(vk)) => self.decrypt_value(vk)?,
+            _ => return Err(Error::MissingViewKey),
+        };
+
+        if value_commitment(opening.0, opening.1) != self.value_commitment {
+            return Err(Error::InvalidData);
         }
```
moCello: **NOTE:** Commitment mismatch reuses `Error::InvalidData`, making a crafted opening indistinguishable from a byte-decode failure or the blinder-conversion failure inside `decrypt_value`.

**Impact:** A wallet cannot distinguish a malformed note from one whose encrypted opening deliberately lies about its commitment — the attack #290 describes, and worth surfacing distinctly (skip vs. warn).

**Fix:** Add a dedicated variant (e.g. `Error::CommitmentMismatch`) and return it here; maintainer judgment call whether the enum addition is worth it before a breaking release.

core/src/note.rs
```diff
@@ -407,37 +411,44 @@ impl Note {
...
-        match (self.note_type, vk) {
-            (NoteType::Transparent, _) => Ok(TRANSPARENT_BLINDER),
-            (NoteType::Obfuscated, Some(vk)) => self
-                .decrypt_value(vk)
-                .map(|(_, value_blinder)| value_blinder),
-            _ => Err(Error::MissingViewKey),
-        }
+        self.value_opening(vk)
+            .map(|(_, value_blinder)| value_blinder)
     }
```
moCello, line 459: **NOTE:** Every `value()`/`value_blinder()` call now pays a two-scalar-mul Pedersen recompute; on the transparent path one of the two is `GENERATOR_NUMS * 0` and avoidable.

**Context:** Callers that fetch both — the common downstream pattern — pay the decrypt and the recompute twice per note; `value_opening` is the single-shot path. The transparent cost lands where `value(None)` runs per note in bulk: the transfer contract's tree insertion and wallet balance scans.

**Fix:** In the transparent arm, compare against `transparent_value_commitment(value)` — same point, half the cost, consistent with `Note::transparent_stealth`. The obfuscated-path cost is the point of the fix and stays.

moCello (commented):
COMMENT — the validation is correct and fully tested, consistent for contract-minted transparent notes, and matches what #290 asked for; nothing gates the merge. One warning and three notes below; the warning's fix is a doc sentence plus a changelog line.

One follow-up that needs no change in this PR: transparent `value(None)` was effectively infallible and is now fallible, so downstream call sites that classify notes with `is_ok`-style checks change meaning for commitment-mismatched notes — rusk's transfer contract classifies zero-value leaves with `note.value(None).is_ok_and(|value| value == 0)`, which now routes a mismatched note into the "not zero" branch instead of classifying it by its encrypted amount. That particular path is already fenced on the node side — `Transaction::phoenix_transparent_output_check` rejects, at mempool admission and again at execution, any Phoenix transparent output that is non-zero or whose zero value does not commit to the canonical zero commitment — so a mismatched transparent output cannot reach the tree through a transaction once that precondition is live. The flip still matters wherever the accessors classify notes outside that fence: the circuit does not constrain `value_enc` until #292. When bumping phoenix-core in rusk, audit `value()`/`value_blinder()` call sites for `is_ok`/`ok()`/`unwrap_or_default` patterns and decide per site whether a mismatch means "skip", "treat as zero", or "reject"; migrate paired accessor calls to a single `value_opening` call while there.

xevisalle (approved):

